PR #28, opened 2022-11-16
fix test_pop_general_success_2 comment

xstarjx:
The nodes should be inner node instead of outer node after popping the tree


alnoki:
Thank you for your contribution! I'll note that for v4 we are replacing the vector-based critbit tree with a table-based AVL/doubly linked list hybrid data structure

PR #35, opened 2022-11-30
…add get_all_market_account_ids_for_user_sdk

yevgenypi:
Add get_all_market_account_ids_for_user_sdk function for querying market account ids with transaction simulation.
This function could significantly reduce the number of requests to get all market account for a user.


alnoki:
Thank you for your contribution to Econia!

As just mentioned outside of GitHub, this feature is covered by #36 , which also adds support for a host of other public getters. Thank you for helping to make Econia more accessible for all.

PR #45, opened 2022-12-07
Add docusaurus docs

BriungRi:
NOTE: I did not carry over the files from `doc/doc-site/api/*.md`


alnoki:
> Looking good, I just have a few minor notes:
> 
> 1. Is there a way to do away with duplicates in `static/img/` and keep relative links to the originals in the repo (e.g. `../../diagrams/ ...`, or substitute for GitHub URLs (analagous to the internal DocGen linking reworks)? Else the asset collections could end up out of sync if someone updates upstream and doesn't update in `static/img`.
> 2. How about we keep the deleted `.md` files in a folder `doc-site/scraps` so they can be referenced as the docs get re-written? For this modification I can commit on before merging or you can commit, whichever you prefer.

@BriungRi ignore the most recent force push, which was a reversion to your most recent commit after an invalid push.

PR #63, opened 2023-01-02
Create devcontainer.json

mehdismh:


alnoki:
@mehdismh what exactly is this for? Econia doesn't have a Docker container set up (not even [`aptos-core`] does at this point, if I'm not mistaken), and any developer workflows you may encounter should be covered by the README's [developer setup] section. Ideally working with Econia would require the minimum amount of installs and dependencies beyond what we already have there.

[`aptos-core`]:   https://github.com/aptos-labs/aptos-core
[developer setup]: https://github.com/econia-labs/econia#developer-setup

PR #71, opened 2023-01-11
Use primary/secondary style for prev/next buttons

BriungRi:
This PR serves as a preview. Needs design input before merging as we are not 100% sure this is the style we want yet.


BriungRi:
Seems like design looks good. Ready for merge.

PR #76, opened 2023-03-09
Setup API

bogberry:


alnoki:
@bogberry looks good, I added in some missing dependencies to the workspace `.toml` file and updated the environment setup.

PR #77, opened 2023-03-09
TypeScript linter and formatter config

bogberry:


alnoki:
@bogberry looks good, I added some TS env init calls in `scripts.sh` and abstracted some of the calls, updated the README and committed the package lock.

PR #78, opened 2023-03-10
run pnpm install in init

bogberry:


alnoki:
Looks good! I just ran the code formatter for the shell script as well.

PR #80, opened 2023-03-15
SQL linter

bogberry:


alnoki:
Since `pip` installs to a global environment, I adapted the formatter to use the same `poetry` env management tool as the other Econia build scripting package. Other than that I just did some alphabetization.

PR #82, opened 2023-03-15
Markets API endpoint

bogberry:


src/rust/types/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub tick_size: u64,
+    pub min_size: u64,
+    pub underwriter_id: u64,
+    pub created_at: DateTime<Utc>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct QueryMarket {
+    pub market_id: BigDecimal,
+    pub base_symbol: Option<String>,
```
alnoki, line 34: Should be some only if base asset is a coin

src/rust/types/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub base_name: Option<String>,
+    pub base_decimals: Option<i16>,
+    pub base_account_address: String,
+    pub base_module_name: String,
+    pub base_struct_name: String,
+    pub quote_symbol: Option<String>,
```
alnoki, line 40: Should never be none since quote asset is always a coin

src/rust/types/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub quote_name: Option<String>,
```
alnoki, line 41: Should never be none, since quote asset it always a coin

src/rust/types/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct QueryMarket {
+    pub market_id: BigDecimal,
+    pub base_symbol: Option<String>,
+    pub base_name: Option<String>,
+    pub base_decimals: Option<i16>,
+    pub base_account_address: String,
+    pub base_module_name: String,
+    pub base_struct_name: String,
```
alnoki, line 39: Should correspond to econia::registry::GenericAsset if a generic asset

src/rust/api/src/routes.rs
```diff
@@ -0,0 +1,114 @@
...
+        types::QueryMarket,
+        r#"
+        select 
+            market_id,
+            base_coins.symbol as base_symbol,
+            base_coins.name as base_name,
+            base_coins.decimals as base_decimals,
+            base_coins.account_address as base_account_address,
+            base_coins.module_name as base_module_name,
+            base_coins.struct_name as base_struct_name,
```
alnoki, line 24: This would be more appropriate as `base_asset`, since the base asset can be either a coin or a `GenericAsset`.

src/rust/types/src/lib.rs
bogberry: Yup `base` on `Market` should be an `Option<Coin>`

src/rust/types/src/lib.rs
bogberry: This is just out of necessity. Basically `MarketRegistrationEvent` only has `TypeInfo`, but the rest of this information has to be retrieved from `CoinInfo`.

I thought it would be better just to have the indexer put nulls in Postgres and have some other task retrieve this information upon a new insert in the markets table. This would require that these columns not be `not null`, and that in turn requires the type here to be `Option<String>`.

bogberry: Still thinking about what the best way to do this would be tbh

qdrs (approved):

PR #83, opened 2023-03-16
switch to brew for installing sqlfluff

bogberry:


alnoki:
@bogberry looks good, I updated the code formatter call to use the CLI version instead of the wrapped poetry call.

PR #120, opened 2023-04-07
SQLFluff config

bogberry:
Adding explicit SQLFluff configurations to avoid formatting rule conflicts between devs


alnoki (changes requested):
I notice that this is modifying all of the jupyter notebooks.

alnoki:
@bogberry It looks like you may have had `jupyter-black` installed in your environment, and this runs automatically whenever black is invoked. I didn't have it, so I added it to the `poetry` config.

PR #148, opened 2023-04-20
ECO-109: TS-SDK

qdrs:


src/typescript/sdk/types/order.d.ts
```diff
@@ -0,0 +1,61 @@
+import { type BCS } from "aptos";
+
+// Order Side
+export enum Side {
+  Ask = 0,
+  Bid = 1,
+}
```
bogberry, line 7: It's better to use string unions rather than numeric enums, because TypeScript type checking doesn't work well with numeric enums. For example, it won't complain if instead of

```ts
txBuilder.placeLimitOrderUserEntry(
  "0x1::aptos_coin::AptosCoin",
  "0x2::test_usdc::TestUSDCoin",
  0n,
  "0x3",
  Side.Bid,
  1000n,
  1000n,
  Restriction.ImmediateOrCancel,
  SelfMatchBehavior.Abort
);
```

You enter

```ts
txBuilder.placeLimitOrderUserEntry(
  "0x1::aptos_coin::AptosCoin",
  "0x2::test_usdc::TestUSDCoin",
  0n,
  "0x3",
  5
  1000n,
  1000n,
  Restriction.ImmediateOrCancel,
  SelfMatchBehavior.Abort
);
```

src/typescript/sdk/src/index.ts
```diff
@@ -1 +1,26 @@
...
+    userAccount: AptosAccount
+  ) {
+    this.econiaAddress = econiaAddress;
+    this.aptosClient = new AptosClient(nodeUrl);
+    this.userAccount = userAccount;
+  }
+
+  public createTx(): EconiaTransactionBuilder {
+    return new EconiaTransactionBuilder(this);
+  }
```
bogberry, line 25: I find this design puzzling. Having a function called `createTx` return an `EconiaTransactionBuilder` seems unintuitive.

src/typescript/sdk/src/tx-builder.ts
```diff
@@ -0,0 +1,345 @@
...
+import { type MaybeHexString } from "aptos";
+
+import { type AdvanceStyle, type Side } from "../types/order";
+import { type Restriction } from "../types/order";
+import { type SelfMatchBehavior } from "../types/order";
+import { type EconiaClient } from ".";
+
+export class EconiaTransactionBuilder {
+  client: EconiaClient;
+  entry?: Types.EntryFunctionPayload;
```
bogberry, line 11: Why have `entry` be an optional field? Why not just have all of the functions below return a `Types.EntryFunctionPayload`?

src/typescript/sdk/src/tx-builder.ts
```diff
@@ -0,0 +1,345 @@
...
+    const signedTx = await this.client.aptosClient.signTransaction(
+      this.client.userAccount,
+      tx
+    );
+    const pendingTx = await this.client.aptosClient.submitTransaction(signedTx);
+    const result = await this.client.aptosClient.waitForTransactionWithResult(
+      pendingTx.hash
+    );
+    return result;
+  }
```
bogberry, line 344: I suppose this function is the reason `entry` is set as a field, but simply having a `submitTx` function on the client that takes in a `Types.EntryFunctionPayload` argument seems like a much cleaner approach.

src/typescript/sdk/src/tx-builder.ts
bogberry, line 1: File names for TypeScript files should be snake case, not kebab case.

bogberry (changes requested):

src/typescript/sdk/src/tx-builder.ts
```diff
@@ -0,0 +1,345 @@
...
+    this.entry = entry;
+    return this;
+  }
+
+  public swapBetweenCoinstoresEntry(
+    base: Types.MoveType,
+    quote: Types.MoveType,
+    marketId: BCS.Uint64,
+    integrator: MaybeHexString,
+    direction: boolean,
```
alnoki, line 191: Can we make this an enum? `ASK` vs `BUY`

alnoki (changes requested):

alnoki (approved):

PR #150, opened 2023-04-20
ECO-110: Reduce market order function to limit order

### LINKED ISSUE #144: Remove place market order entry function
## Description

The current implementation of the entry functions used for placing market orders include arguments such as:

- `limit_price`
- `min_base`
- `max_base`
- `min_quote`
- `max_quote`

which do not conform to industry standards and a commonly accepted definition of a market order. Typically, only a size will be specified for a market order, and there should be no guarantees on what price it fills at or what amount of coins will be available after it has been filled.

For example, the definition given on the introductory article on [investor.gov](https://www.investor.gov/introduction-investing/investing-basics/how-stock-markets-work/types-orders) states

> A market order is an order to buy or sell a security immediately. This type of order guarantees that the order will be executed, **but does not guarantee the execution price**.

By definition, any order which has a limit price above or below it will not execute is a limit order, not a market order. It does not make sense to provide this sort of granularity in a market order, since this is not the purpose a market order is intended to serve.

## Possible solutions

To avoid confusion, we should either 
- remove the market order functions entirely, or
- replace them with function that does not take in the above arguments.

If we remove the market order functions entirely, users would submit immediate-or-cancel limit orders with either a size of 0 or a size of `U64_MAX`, depending on whether it is a bid or an ask. This would allow us to make this fix purely by removing code, not adding anything, and it may allow us to make this change without any additional audits.

Alternatively, we could replace these functions with updated versions that do not take in these arguments. This is fairly simple to implement, and if we are going to audit the diff between the audited version and a fix for #143 anyways, we might as well for convenience's sake.

qdrs:
closes #144

### DIFF
--- a/src/move/econia/sources/market.move
+++ b/src/move/econia/sources/market.move
@@ -1173,11 +1173,7 @@ module econia::market {
         market_id: u64,
         integrator: address,
         direction: bool,
-        min_base: u64,
-        max_base: u64,
-        min_quote: u64,
-        max_quote: u64,
-        limit_price: u64,
+        size: u64,
         self_match_behavior: u8,
         custodian_capability_ref: &CustodianCapability
     ): (
@@ -1191,11 +1187,7 @@ module econia::market {
             registry::get_custodian_id(custodian_capability_ref),
             integrator,
             direction,
-            min_base,
-            max_base,
-            min_quote,
-            max_quote,
-            limit_price,
+            size,
             self_match_behavior)
     }
 
@@ -1214,11 +1206,7 @@ module econia::market {
         market_id: u64,
         integrator: address,
         direction: bool,
-        min_base: u64,
-        max_base: u64,
-        min_quote: u64,
-        max_quote: u64,
-        limit_price: u64,
+        size: u64,
         self_match_behavior: u8
     ): (
         u64,
@@ -1231,11 +1219,7 @@ module econia::market {
             NO_CUSTODIAN,
             integrator,
             direction,
-            min_base,
-            max_base,
-            min_quote,
-            max_quote,
-            limit_price,
+            size,
             self_match_behavior)
     }
 
@@ -1803,16 +1787,12 @@ module econia::market {
         market_id: u64,
         integrator: address,
         direction: bool,
-        min_base: u64,
-        max_base: u64,
-        min_quote: u64,
-        max_quote: u64,
-        limit_price: u64,
+        size: u64,
         self_match_behavior: u8
     ) acquires OrderBooks {
         place_market_order_user<BaseType, QuoteType>(
-            user, market_id, integrator, direction, min_base, max_base,
-            min_quote, max_quote, limit_price, self_match_behavior);
+            user, market_id, integrator, direction,
+            size, self_match_behavior);
     }
 
     #[cmd]
@@ -2225,7 +2205,10 @@ module econia::market {
         bool
     ) {
         // Assert price is not too high.
-        assert!(limit_price <= HI_PRICE, E_PRICE_TOO_HIGH);
+        if (limit_price > HI_PRICE) {
+            limit_price = HI_PRICE;
+        };
+
         // Taker buy fills against asks, sell against bids.
         let side = if (direction == BUY) ASK else BID;
         let (lot_size, tick_size) = (order_book_ref_mut.lot_size,
@@ -2530,9 +2513,11 @@ module econia::market {
     ) acquires OrderBooks {
         // Assert valid order restriction flag.
         assert!(restriction <= N_RESTRICTIONS, E_INVALID_RESTRICTION);
-        assert!(price != 0, E_PRICE_0); // Assert nonzero price.
+        // assert!(price != 0, E_PRICE_0); // Assert nonzero price.
         // Assert price is not too high.
-        assert!(price <= HI_PRICE, E_PRICE_TOO_HIGH);
+        if (price > HI_PRICE) {
+            price = HI_PRICE;
+        };
         // Get user's available and ceiling asset counts.
         let (_, base_available, base_ceiling, _, quote_available,
              quote_ceiling) = user::get_asset_counts_internal(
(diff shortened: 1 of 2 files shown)


src/move/econia/sources/market.move
```diff
@@ -2530,9 +2513,11 @@ module econia::market {
     ) acquires OrderBooks {
         // Assert valid order restriction flag.
         assert!(restriction <= N_RESTRICTIONS, E_INVALID_RESTRICTION);
-        assert!(price != 0, E_PRICE_0); // Assert nonzero price.
+        // assert!(price != 0, E_PRICE_0); // Assert nonzero price.
         // Assert price is not too high.
-        assert!(price <= HI_PRICE, E_PRICE_TOO_HIGH);
+        if (price > HI_PRICE) {
+            price = HI_PRICE;
+        };
```
alnoki, line 2520: If there is no check against the price being 0, then a user will be able to place a maker order on the book with price 0.

If the concern here is to make sure that market orders can fill adequately, then we could easily define `MIN_PRICE = 1` and have market order functions pass that in.

Also per above it looks like we don't strictly need this and so I'd advocate for minimum refactor, and also the doc comment hasn't been updated to reflect the changes.

@bogberry your thoughts?

src/move/econia/sources/market.move
```diff
@@ -2964,72 +2958,30 @@ module econia::market {
...
-            (base_withdraw - base_traded);
-        // Deposit assets back to user's market account.
-        user::deposit_assets_internal<BaseType, QuoteType>(
-            user_address, market_id, custodian_id, base_deposit,
-            optional_base_coins, quote_coins, underwriter_id);
-        // Return base and quote traded by user, fees paid.
+        let price: u64 = if (direction == ASK) {
+            0
+        } else {
+            MAX_POSSIBLE
```
alnoki, line 2971: Market orders and swap APIs accept `direction` = `BUY` or `SELL` so I suggest a ternary of the form `let side = if (direction == BUY) BID else ASK`.

Then for the price ternary `let price = if (direction == ASK) MIN_PRICE else HI_PRICE`.

With `MIN_PRICE = 1` as I suggest above, this will require minimum refactors in the limit order function and in the match function.

src/move/econia/sources/market.move
```diff
@@ -3096,15 +3048,15 @@ module econia::market {
...
         min_quote: u64,
         max_quote: u64,
-        base_available: u64,
-        base_ceiling: u64,
-        quote_available: u64,
-        quote_ceiling: u64
+        _base_available: u64,
+        _base_ceiling: u64,
+        _quote_available: u64,
+        _quote_ceiling: u64
```
alnoki, line 3059: Why are you nulling these? The limit order and the swap functions rely on this.

src/move/econia/sources/market.move
```diff
@@ -2566,14 +2551,23 @@ module econia::market {
...
+        if (ticks > (HI_64 as u128)) {
+            ticks = (HI_64 as u128);
+        };
         // Assert corresponding tick amount fits in a u64.
         assert!(ticks <= (HI_64 as u128), E_SIZE_PRICE_TICKS_OVERFLOW);
         // Calculate amount of quote required to fill size at price.
         let quote = ticks * (order_book_ref_mut.tick_size as u128);
+        if (quote > (HI_64 as u128)) {
+            quote = (HI_64 as u128);
+        };
```
alnoki, line 2570: These should get covered by the range checks below. I advocate the minimum refactor possible and I don't think we need these.

@bogberry your thoughts?

src/move/econia/sources/user.move
```diff
@@ -1751,8 +1751,9 @@ module econia::user {
             (size as u128) * (market_account_ref_mut.lot_size as u128);
         // Calculate ticks to fill order.
         let ticks = (size as u128) * (price as u128);
-        // Assert ticks to fill order is not too large.
-        assert!(ticks <= (HI_64 as u128), E_TICKS_OVERFLOW);
+        if (ticks > (HI_64 as u128)) {
+            ticks = (HI_64 as u128)
+        };
```
alnoki, line 1756: Silent reassignment without feedback here will create an accounting error against the order book, which maintains its own copy of state. I suggest this be removed.

src/move/econia/sources/user.move
```diff
@@ -1772,12 +1773,14 @@ module econia::user {
...
+        if (in_fill + (*in_ceiling_ref_mut as u128) > (HI_64 as u128)) {
+            in_fill = (HI_64 as u128) - (*in_ceiling_ref_mut as u128);
+        };
         // Assert enough outbound asset to cover the fill, which also
         // ensures outbound fill amount does not overflow.
-        assert!((out_fill <= (*out_available_ref_mut as u128)),
-                E_NOT_ENOUGH_ASSET_OUT);
+        if (out_fill > (*out_available_ref_mut as u128)) {
+            out_fill = (*out_available_ref_mut as u128);
+        };
```
alnoki, line 1783: Same here on the silent reassingment, this can create an accounting error against the order state on the order book.

src/move/econia/sources/user.move
```diff
@@ -2240,7 +2243,9 @@ module econia::user {
                 &mut market_account_ref_mut.quote_ceiling
             ) else abort E_ASSET_NOT_IN_PAIR;
         // Assert enough asset available for withdraw.
-        assert!(amount <= *available_ref_mut, E_WITHDRAW_TOO_LITTLE_AVAILABLE);
+        if (amount > *available_ref_mut) {
+            amount = *available_ref_mut
+        };
```
alnoki, line 2248: I believe this can result in an accounting error when routing assets between counterparties.

alnoki (changes requested):

alnoki:
@qdrs I notice there are several refactors involving reassignments that are not necessary, with some that could have potentially-adverse consequences.

It appears that the minimum change set required for #144  is simply to change the internals of the `place_market_order` function and add in a `MIN_PRICE = 1`, and so I suggest this approach.

Also, do you mind removing the `closes #144` tag at the top? There will still need to be documentation updates to complete the revisions.

alnoki:
@qdrs @bogberry See #151 for a much simpler refactor.

PR #178, opened 2023-05-10
Update Header

BriungRi:
Depends on [177](https://github.com/econia-labs/econia/pull/177)

### DIFF
--- a/src/typescript/frontend/shell.nix
+++ b/src/typescript/frontend/shell.nix
@@ -9,5 +9,6 @@ pkgs.mkShell {
   shellHook = ''
     alias pd='pnpm run dev'
     alias pb='pnpm run build'
+    alias pl='pnpm run lint'
   '';
 }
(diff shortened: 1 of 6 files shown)


bogberry:
The Next.js <Link> component is only meant to be used for internal links, not external ones. I changed NavItem to simply use an <a> tag for external links.

bogberry:
I also prefer having the `hover:text-purple` class only when the NavItem is not active, since if you have it on both, you get a flashing effect where after you click a link, the color of the NavItem briefly changes to the active color, but then quickly changes to purple because your mouse is still hovering over the element.

bogberry:
Also super minor but you don't need `rel="noopener"` these days
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#security_and_privacy

bogberry (approved):

bogberry (approved):

PR #180, opened 2023-05-10
Refactor WalletSelector => ConnectedButton

BriungRi:
Depends on #179 

`ConnectedButton` is a useful abstraction is because it means we can reuse it for places like the primary Swap action, Faucet action.

Though, consider moving the Modal to the `useWallet` context so that we aren't unnecessarily including multiple of the same wallet connection Modal.

This diff also moves the `Connect Wallet` button into the header.

### DIFF
--- a/src/typescript/frontend/shell.nix
+++ b/src/typescript/frontend/shell.nix
@@ -9,5 +9,6 @@ pkgs.mkShell {
   shellHook = ''
     alias pd='pnpm run dev'
     alias pb='pnpm run build'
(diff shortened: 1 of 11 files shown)


bogberry (approved):

bogberry:
The `useWallet` context comes from the `@manahippo/aptos-wallet-adapter` library, so I don't think moving the modal is an option unless we want to go through the trouble of forking our own version?

Otherwise looks good. Agreed about `ConnectedButton` being a useful abstraction.

PR #183, opened 2023-05-11
Change color of icons when user hovers over them

bogberry:


chiumax:
merged with https://github.com/econia-labs/econia/pull/182
closing.

PR #184, opened 2023-05-11
Add trading view as submodule

bogberry:


BriungRi:
for charting lib, thoughts on us making a private fork that upstreams to the official one, so that our team on GH all has access to the library?

also, is there a good way for our frontend to work for others who don't have access to the charting lib? Or will they always need to get their own access the charting lib?

PR #193, opened 2023-05-14
Add scaffolding for MarketOrderEntry

BriungRi:
Depends on #192


src/typescript/frontend/src/components/trade/OrderEntry/MarketOrderEntry.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <div className="mx-4">
+        <OrderEntryInput
+          value={price}
+          onChange={setPrice}
+          startAdornment="MARKET PRICE"
+          endAdornment={marketData.quote.symbol}
+          type="number"
+          placeholder="0.00"
+          disabled
+        />
```
bogberry, line 28: I don't think the market order form should have a market price component. Average fill price will depend on the size of the order and I don't think it's worth trying to calculate it on the fly depending on what the user enters in the size input.

bogberry (approved):

PR #208, opened 2023-05-22
Move ConnectWallet modal to its own context

BriungRi:
Depends on #206


bogberry:
Also, this isn't part of this PR, but the button on the order entry form should not submit the form when it is a "connect wallet" button, and not "buy" or "sell." Submitting the form triggers input validation, which shouldn't be necessary for someone clicking to connect their wallet.

juicedlime:
> Isn't that already the case? The `Connect Wallet` button isn't a `type="submit"`.

Doesn't seem like it, maybe you need `e.preventDefault()`

bogberry (approved):

PR #217, opened 2023-05-24
Fix migration

qdrs:


bogberry (changes requested):
`register_market_trigger` is set so that `register_market()` is run prior to the insertion on `market_registration_events`, so the row on the `markets` table be inserted beforehand.

qdrs:
> `register_market_trigger` is set so that `register_market()` is run prior to the insertion on `market_registration_events`, so the row on the `markets` table be inserted beforehand.

this is now, fixed, it required setting the market_registration_events foreign key as deferrable

PR #223, opened 2023-05-26
Show market account balance in deposit/withdraw modal

BriungRi:
Also refactor types into respective files and move TypeTag into utils/


bogberry (approved):
Looks good!

PR #226, opened 2023-05-27
Mc/market select

chiumax:
need to test many rows
wip

todo:
optional exit fn for base modal - done
make sure empty table has same widths - done

filter by recognized - done BUT USING MOCK DATA

image


src/typescript/frontend/src/components/BaseModal.tsx
```diff
@@ -9,8 +9,9 @@ export const BaseModal: React.FC<
     open: boolean;
     onClose: () => void;
     onBack?: () => void;
+    showCloseButton?: boolean;
   }>
-> = ({ open, onClose, onBack, children }) => {
+> = ({ open, onClose, onBack, showCloseButton = true, children }) => {
```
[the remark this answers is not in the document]
chiumax: disagree, in the case of the design for `StatsBar` i still would need `onClose` to set modal open state to false. but i don't necessarily want the close button to show up

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
```diff
@@ -1,116 +1,490 @@
...
+                    <td colSpan={7}>
+                      <div className="flex h-[150px] flex-col items-center justify-center text-sm font-light uppercase text-neutral-500">
+                        No markets to show
+                      </div>
+                    </td>
+                  </tr>
+                ) : (
+                  table.getRowModel().rows.map((row) => (
+                    <tr
+                      className="h-24 min-w-[780px] cursor-pointer px-6 text-left font-roboto-mono text-sm text-white hover:outline hover:outline-1 hover:outline-neutral-600 [&>th]:font-light"
```
bogberry, line 242: The hover border and modal border blend into one another, so it looks kind of strange and unclear what market is being selected, especially with the last market on the list. Maybe we can come up with something better?

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
```diff
@@ -1,116 +1,490 @@
...
+    compactDisplay: "short",
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 1,
+  });
+  return (
+    <div>
+      <div className={`inline-block min-w-[8em] text-sm`}>
+        {formatter.format(volume).replace("K", "k")} {baseAsset}
+      </div>
+      <div className={`inline-block min-w-[6em] text-neutral-500`}>$1.5M</div>
```
bogberry, line 334: I think volume will have to be base currency / quote currency. If we want to give volume in USD, there's a separate discussion to be had about where the USD price data comes from

src/typescript/frontend/src/components/BaseModal.tsx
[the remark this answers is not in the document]
bogberry: Yup I think that's reasonable

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
```diff
@@ -1,116 +1,490 @@
...
+        alt="market-icon-pair"
+        width={40}
+        height={40}
+        className="absolute z-10 aspect-square w-[30px] min-w-[30px] translate-x-1/2 md:min-w-[40px]"
+      ></ImageWithFallback>
+    </div>
+  );
+};
+
+const getLang = () => {
```
bogberry, line 451: Probably should move number formatter / get locale function to a separate utils file, since I assume that’ll be used in a lot of places

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
```diff
@@ -1,116 +1,490 @@
...
+                    })}
+                  </tr>
                 ))}
-              </tr>
-            ))
-          )}
-        </tbody>
-      </table>
+              </thead>
+              <tbody>
```
bogberry, line 217: Modal should be fixed height and have the table body overflow on the y-axis when there are too many markets to fit on screen. Currently, the entire modal just scrolls, which isn't ideal

bogberry (changes requested):

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
chiumax: good point, will bring up with @xorkiki

chiumax: ah just saw slack, nvm

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
chiumax: blocked: waiting for input from others

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
chiumax: yeah expected behavior is to have the table scroll imo

chiumax: will likely just set a minheight, but need to account for mobile screensize if height is < modal height

PR #241, opened 2023-05-30
Mc/consolidate util

chiumax:
consolidate common util functions 
get coinlist images in market select


chiumax:
> Why are there changes for the pnpm-lock.yaml files?
idk, is it bad to run `pnpm install` after a merge?

PR #252, opened 2023-06-01
Orderbook select populate order

chiumax:
when a row on orderbook is selected, auto populate order entry
row 22 
row 24
on bug spreadsheet


chiumax:
would this be useful for future components?

bogberry (approved):

PR #255, opened 2023-06-02
Update metadata and add favicon to API docs page

bogberry:


bogberry:
That's a good idea. I tried setting the frontend favicon to be the source, and having both documentation sites use symlinks, and it seems to be fine at least locally.

Let me try merging this to make sure it doesn't cause issues for any of our deployments either.

bogberry:
Seems to cause no issues in deployment either

PR #258, opened 2023-06-02
Layout when order switch

chiumax:
this was the only way i could think of to consistently keep a "set" chart height.
Because the chart inherently does not have any height, it's height depended on the limit order component.
to keep the chart a consistent size, added a min height according to spec.

though now I think the page will overflow vertically regularly, so take a look into this w/ kiki
row 28 in bugs spreadsheet


bogberry (approved):

bogberry:
Yup I think it makes sense to have a set height for the depth chart. We should probably change the order form so the size doesn't change and cause layout shifts when you select limit / market, but that can be a separate PR.

PR #261, opened 2023-06-05
Use atomically reference counted struct for app state in API

bogberry:


elliottdehn:
Are you able to say why this is necessary? Thanks.

bogberry:
This reduces the number of times data needs to be copied elsewhere in the code.

PR #263, opened 2023-06-05
Decimals

chiumax:
make sure numbers in orderbook, depth chart, stats bar are all displayed to spec
https://econia.dev/overview/orders/#units-and-market-parameters


chiumax:
what if base is null?

PR #264, opened 2023-06-05
Market Select Modal consistent sizing

chiumax:
thinking about whether to size the market select using:
- vh
- or just to spec with px?

how to communicate to the user the modal is scrollable / there is more rows?
edit: adjusting height so if there are more rows, the bottom one is cut off

update:
unable to fix sizing without styling going off.
apparently it's not possible to stack `overflow-y: auto` with `overflow-x: visible`

potential solutions:
- position absolute, and position remaining elements hard code
- hardcode the height of the table
- pass optional padding prop to `BaseModal` and then add padding manually to the market selelct modal (doing this as of now)


src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
```diff
@@ -104,13 +104,13 @@ export const SelectMarketContent: React.FC<{
     getCoreRowModel: getCoreRowModel(),
   });
   return (
-    <div className="flex w-full flex-col items-center">
+    <div className="flex max-h-[655px] min-h-[655px] w-full flex-col items-center overflow-y-hidden">
```
bogberry, line 107: Is there a reason you need to set min and max here, and not simply set the height to 655px?

src/typescript/frontend/src/components/BaseModal.tsx
```diff
@@ -32,7 +41,9 @@ export const BaseModal: React.FC<
         <div className="fixed inset-0 overflow-y-auto">
           <div className="flex min-h-full items-center justify-center p-4 text-center">
             <Dialog.Panel
-              className="w-full max-w-4xl transform border border-neutral-600 bg-black p-6 align-middle shadow-xl transition-all"
+              className={`w-full max-w-4xl transform border border-neutral-600 bg-black ${
+                disablePadding === true ? "" : "p-6"
```
bogberry, line 45: I'm a bit unclear about why this `disablePadding` prop needs to exist. Is there some styling on the deposit/withdraw form that can't be achieved by applying styles to the children of this component?

bogberry (changes requested):

src/typescript/frontend/src/components/trade/DepositWithdrawModal/SelectMarketContent.tsx
chiumax: some weird css behavior when a max and min is not set

src/typescript/frontend/src/components/BaseModal.tsx
chiumax: yes, when each row is hovered, the entire row from edge to edge should highlight. but there is no way to extend the width of the rows to somehow have negative padding to "negate" padding form the parent.

bogberry: Oh no my suggestion would be to default to 0 padding and add a margin or an extra div with padding to the child components when you need it

bogberry: I actually made this change upstream so nothing to do in this PR

bogberry (approved):

PR #265, opened 2023-06-05
ECO-174: orderbook websocket

chiumax:
```
/**
   * assumption made is there is always 60 entries in the orderbook visible at any time.
   * things to account for:
   *
   *
   * 1. if price level size is 0, then remove it from the orderbook. need to think about how to update range to account for this.
   * 2. if price level does not exist, and lies between the range, then add it to the orderbook. remove out of range price levels and update range
   * 3. account for precision???
   */
```

i guess basically, do we care about how many price levels we show on the frontend? 
or just insert / remove updates accordingly?
- new price level, add
- price level with 0 volume, remove

and then refetch the whole orderbook every once in a long while? idk


chiumax:
todo use api env or create new one

PR #275, opened 2023-06-07
Eco 133: add sorting to orders table

chiumax:
better styling for sorting indicator?
also, which columns should have sorting?

### DIFF
--- a/src/typescript/frontend/src/components/trade/OrdersTable.tsx
+++ b/src/typescript/frontend/src/components/trade/OrdersTable.tsx
@@ -5,12 +5,15 @@ import {
   flexRender,
   getCoreRowModel,
   useReactTable,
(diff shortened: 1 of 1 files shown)


chiumax:
yea i agree, this still needs some work, making it a draft again

bogberry:
- I agree with @BriungRi, the column headers should be `cursor: pointer` and the text should not be highlighted.
- Probably should make it possible for the user to sort on every column.
- The sort icons should be more prominent. Discuss with @xorkiki?
- As @BriungRi mentioned, sorting by a column should not cause a layout shift.

bogberry:
I've made this into a draft

chiumax:
closing in favor of: https://github.com/econia-labs/econia/pull/378

PR #276, opened 2023-06-07
Eco 134: orderbook price level Flash animation when updated

chiumax:
before the last commit, with how animations are being applied, the flash animation will get interrupted if there are "simultaneous" updates. this is more apparent when setting the flash duration to be longer.
please take a look, it might be expensive for the frontend to do this?

depends on #265 

==== not important ===
 
Some thoughts + considerations:
1. maybe the animations are short enough to not notice? I did testing myself and personally could tell that it looked a bit glitchy.
2. the solution to this would be

```
/**
 * putting down some thoughts here
 * initially thought itd be better to extend PriceLevel, but then realized that there's no good way to remove the didUpdate prop without really keeping track of the orderbook
 * thinking it might be better to extend orderbook to keep track of the most updated pricelevel
 * ---
 * issue is if the animation is 2s long in duration
 * and updates come in 1s within each other, then that means that the prev css class will be removed before it ends
 * but does this necessarily cancel the css animation?
 */

/**
     * 1. update animation - this works
     * 2. same level getting updated twice - this works
     * 3. levels getting updated in quick succession before animation ends - this does not work yet
     * 4. same level getting updated in quick succession before animation ends - this works
     */
```

depends on orderbook-websocket

### DIFF
--- a/src/typescript/frontend/src/components/OrderBook.tsx
+++ b/src/typescript/frontend/src/components/OrderBook.tsx
@@ -2,12 +2,12 @@ import { Listbox } from "@headlessui/react";
 import { CheckIcon, ChevronDownIcon } from "@heroicons/react/20/solid";
 import { useEffect, useMemo, useRef, useState } from "react";
 
+import { useOrderEntry } from "@/contexts/OrderEntryContext";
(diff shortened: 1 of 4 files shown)


chiumax:
closing in favor of #352

PR #279, opened 2023-06-12
ECO-172: Change TradingView default colors to match Econia color scheme

bogberry:


bogberry:
> Consider separating changes like `aggregator` into a separate diff. I don't think it's necessary here?

Whoops, I thought I was just merging dev there, but I think my local dev branch was out of sync. Reverted the aggregator changes in this PR.

bogberry:
Fixed

PR #282, opened 2023-06-13
Eco 147 + Eco 146: toast on order cancel / eviction

chiumax:
idk how to test this


src/typescript/frontend/src/hooks/useOrderStatusNotif.tsx
```diff
@@ -0,0 +1,87 @@
...
+    {
+      keepPreviousData: true,
+      refetchOnWindowFocus: false,
+      refetchInterval: 10 * 1000,
+    }
+  );
+
+  useEffect(() => {
+    // todo update env here, api url in env has https:// prefix
+    const websocket = new WebSocket(`wss://dev.api.econia.exchange/ws`);
```
bogberry, line 43: I prefer having the WebSocket connection be defined in the page, and storing it in a ref, since when you put it in a hook it's really easy to accidentally end up in a situation where you're running the `useEffect` hook twice and accidentally creating multiple WebSocket connections.

src/typescript/frontend/src/hooks/useOrderStatusNotif.tsx
```diff
@@ -0,0 +1,87 @@
...
+    websocket.onmessage = (event) => {
+      const data = JSON.parse(event.data).data;
+
+      if (!data) return;
+
+      // transaction hash here?
+      if (data.order_state === "evicted") {
+        toast.info(
+          `Order of ${data.order_size} ${selectedMarket?.base?.name} for ${price.data} ${selectedMarket?.quote.name} is evicted`
+        );
```
bogberry, line 67: I think we'll probably want at least an order ID so that the user can uniquely identify the order given the toast text.

Transaction hash might be useful, but might be a lot to fit in to the toast body? Probably worth experimenting with this

chiumax:
closing, looks resolved in: https://github.com/econia-labs/econia/commit/63e81db8eb7a6c9c7e704a1905939103f6f437a7

PR #284, opened 2023-06-13
Eco 165: Required to be switched to all caps + positioning 

chiumax:


bogberry:
I thought I'd just use the `uppercase` tailwind class here, since typing out some of the other validation errors you could get here might get kinda annoying. Not a big deal either way though.

Positioning changes look good to me

bogberry (approved):

PR #285, opened 2023-06-13
Pin indexer submodule branch

alnoki:


bogberry (changes requested):
This should not be added, as git submodules are pinned to a specific commit, not the latest commit of a branch.

PR #291, opened 2023-06-20
Handle WebSocket API price level updates and update React Query cache

bogberry:
- Handles price level updates from the WebSocket API
- Adds price level type to correspond to schema used by backend
- Updates React Query cache using WebSocket data for orderbook state

Depends on #289.


src/typescript/frontend/src/components/OrderbookTable.tsx
```diff
@@ -69,12 +68,22 @@ const Row: React.FC<{
...
+}) {
   const [precision, setPrecision] = useState<Precision>(precisionOptions[0]);
-  const { data, isFetching, isLoading } = useOrderBook(
-    marketData.market_id,
-    precision
-  );
+  // const { data, isFetching, isLoading } = useOrderBook(
+  //   marketData.market_id,
+  //   precision
+  // );
```
alnoki, line 86: Can we avoid putting commented out code in here?

src/typescript/frontend/src/pages/trade/[market_name].tsx
```diff
@@ -202,14 +207,100 @@ export default function Market({ allMarketData, marketData }: Props) {
...
           // TODO
         }
       } else {
         // TODO
       }
     };
-  }, [marketData?.market_id, account?.address]);
+  }, [marketData, account?.address, queryClient]);
+
+  // TODO update to include precision when backend is updated
```
Comment (alnoki) [inline src/typescript/frontend/src/pages/trade/[market_name].tsx:288]: Can we tag an issue here, or ideally start an issue so that we don't have a floating TODO?

alnoki (changes requested):

src/typescript/frontend/src/components/OrderbookTable.tsx
bogberry: Removed

src/typescript/frontend/src/pages/trade/[market_name].tsx
Comment (bogberry) [reply src/typescript/frontend/src/pages/trade/[market_name].tsx]: Done

PR #312, opened 2023-06-29
Add backwards-compatible taker order ID support

alnoki:


alnoki:
# Summary of changes

* Deprecates `TakerEvent` and `OrderBook.taker_events` in lieu of `FillEvent` and `FillEventHandles`.
* Updates order book counter to increment for each order that is submitted (incremented in `match()` and in `place_limit_order()`).
* Modifies `match()` signature to accept as an argument resource account address, used for lookup on `FillEventHandles`.

# Functional considerations

Compared with the deprecated `TakerEvent` structure, `FillEvent` adds an `order_book_counter` field, which is identical across all `FillEvent` emissions for a single order. Moreover, when a limit order fills as both a maker and as a taker, `FillEvent.order_book_counter` fields are identical to the counter encoded in the `MakerEvent.market_order_id`. This correspondence allows for `FillEvents` to be linked with a market order ID for a limit order that fills partially as a taker and as a maker.

Since fill events do not insert to the order book, they do not generate an "AVL queue access key" (essentially a pointer into order book memory), and as such they do not have a market order ID (which contains an AVL queue access key). In the case of a limit order that partially takes and partially makes, it inserts to the order book (maker) *after* filling (taker), hence the eventual market order ID is not known at the time of `FillEvent` emission.

See `market::get_market_order_id_counter()`.

In terms of post-processing this means that, for a limit order that fills as a taker then as a maker, `FillEvents` can be linked with a `MakerEvent` by comparing `FillEvent.order_book_counter` fields (identical across all fills from a single match) against the counter encoded in the corresponding `MakerEvent.market_order_id`.

As for post-processing, here there are two (perhaps more) potential options:

## Create `FillEvent` market order ID upon insertion

When inserting a `FillEvent` to a database, first check if any `MakerEvents` emitted on the market up to the current ledger version number have a market order ID that encodes the corresponding `FillEvent.order_book_counter`: there will be at most one match, which can then be treating as the corresponding market order ID for all such `FillEvents`. In the case of no match, the market order ID recorded for the corresponding `FillEvent`(s) be generated by simply bitshifting the counter into a `u128`. 

See `market::get_market_order_id_counter()`.

## Correspond `FillEvent`(s) with market order ID upon query

When inserting a `FillEvent` to a database, simply insert the order book counter as-is from the event stream. Then, if a client submits a query of the form `get_all_potential_fills_associated_with_market_order_id(market_order_id)`, extract the counter from the market order ID and look up all `FillEvents` that have the counter value.

See `market::get_market_order_id_counter()`.

# Checks

* f01cf2f Verified as an on-chain upgrade against [v4.0.2-audited](https://github.com/econia-labs/econia/releases/tag/v4.0.2-audited) (within a v2 multisig) at https://aptos-explorer.netlify.app/account/0xdcdf825b41195bb6d5b2b2b95913e344e3714b960ad6fd43b3d20613be0915a3/transactions?network=testnet
* Code tested to 100% coverage

PR #313, opened 2023-06-30
ECO-200: add fills by market and account endpoint

bogberry:


src/rust/db/examples/seed_db.rs
```diff
@@ -58,6 +58,21 @@ fn place_random_orders(
...
+                side,
+                market_order_id: &id.into(),
+                maker: "0x123",
+                custodian_id: None,
+                size: &BigDecimal::from_u64(size >> 2).unwrap(),
+                price: &BigDecimal::from_u64(price).unwrap(),
+                time: Utc::now(),
+            },
+        )
+        .unwrap();
```
alnoki, line 75: It looks like we might deprecate taker events in lieu of fill events per #315

bogberry: Yup. For now let me just remove this file from this PR.

alnoki (approved):

PR #315, opened 2023-06-30
Add FillEvent with common market order ID

alnoki:
# Summary of changes

* Deprecates `TakerEvent` and `OrderBook.taker_events` in lieu of `FillEvent` and `FillEventHandles`.
* Updates order book counter to increment for each order that is placed (incremented in `match()` and in `place_limit_order()`).
* Modifies `match()` function to defer `FillEvent` emission in the case of a limit order.
* Adds view function (`did_order_post()`) for determining if a market order ID corresponds to an order that posted to the book.
* Enforces common market order ID across all `FillEvent`(s) when an order results in multiple fills, which is common to the `MakerEvent` market order ID in the case of a limit order that crosses the spread and fills as a taker before posting as a maker.
* Implemented as a backwards-compatible package upgrade.


alnoki:
Note that bac3bf24 has been pushed to `ecoina-labs/econia/fill-events`, yet it is not showing up in this PR

It is only a changelog update, so I am not going to proceeed with this PR as is. I assume GitHub is facing a temporary glitch where new commits on top of a PR are not rendered in the conversation history.

src/move/econia/doc/market.md
```diff
@@ -11,9 +11,10 @@ order book entry is added under the resource account at a new market
 ID.
 
 Once a market is registered, signing users and delegated custodians
-can place limit orders on the book as makers, takers can place
-market orders or swaps against the order book, and makers can cancel
-or change the size of any outstanding orders they have on the book.
+can place limit orders on the book as makers and/or as takers,
+takers can place market orders or swaps against the order book, and
+makers can cancel or change the size of any outstanding orders they
+have on the book.
```
bogberry, line 17: This wording seems strange to me. There is no sense in which any given user is a maker or a taker—any given order can be a maker order or a taker order, and the user will be a maker or a taker with regard to that specific order, but that doesn't describe their relationship to the market itself.

src/move/econia/sources/market.move
```diff
@@ -2766,13 +2903,16 @@ module econia::market {
                         fill_size, complete_fill, optional_base_coins,
                         quote_coins, fill_size * lot_size,
                         ticks_filled * tick_size);
-                // Get taker events handle.
-                let taker_handle = &mut order_book_ref_mut.taker_events;
-                // Emit corresponding taker event.
-                event::emit_event(taker_handle, TakerEvent{
-                    market_id, side, market_order_id, maker, custodian_id:
-                    maker_custodian_id, size: fill_size, price});
```
bogberry, line 2774: Backwards compatibility means that anyone who had built anything on top of a previous version of this code would not have to make any changes for their code to keep on running as expected.

As we are no longer emitting taker events, and any system that relied on taker events would have to make the switch over to fills events, this change is not backwards compatible.

src/move/econia/sources/market.move
```diff
@@ -3089,13 +3238,31 @@ module econia::market {
...
                 size - (base_traded / order_book_ref_mut.lot_size);
-        }; // Done with optional matching as a taker across the spread.
-        // Return without market order ID if immediate-or-cancel or if
-        // remaining size to fill after matching does not meet minimum
-        // size requirement for market.
+            // Extract fill event queue packed by matching engine.
+            option::extract(&mut fill_event_queue_option)
+        } else { // If spread not crossed (matching engine not called):
+            // Increment order counter.
+            order_book_ref_mut.counter = order_book_ref_mut.counter + 1;
```
bogberry, line 3245: Comment simply repeats code

bogberry (changes requested):

src/move/econia/doc/market.md
alnoki: Tagging #321 to incorporate refactors on that iteration

src/move/econia/sources/market.move
alnoki: @bogberry Good point, backwards-compatibility is only in the sense that the contract itself will pass a bytecode verifier upgrade check, required to deploy to the same address on mainnet

Tagging #321 to incorporate refactors on that iteration

src/move/econia/sources/market.move
alnoki: Tagging #321 to incorporate refactors on that iteration

src/move/econia/sources/market.move
bogberry: Yes, so it's not really backward-compatible. Aptos simply checks the struct declarations and function signatures

bogberry: Yes, so it's not really backward-compatible. Aptos simply performs a minimal check where the struct declarations and function signatures are compared to catch any changes that are obviously not backwards compatible, as far as I'm aware. This is in no way a guarantee of backwards compatibility, and since we know that this in fact isn't a backwards compatible change, we should stop describing it as such.

alnoki:
Closing in lieu of more comprehensive #321

PR #321, opened 2023-07-03
Refactor events/logic for indexing support

alnoki:


alnoki:
The most recent commit adds on top of #315 draft structures to address recent discussions about user- and market-level indexing.

src/move/econia/sources/market.move
```diff
@@ -2766,13 +2955,16 @@ module econia::market {
...
-                event::emit_event(taker_handle, TakerEvent{
-                    market_id, side, market_order_id, maker, custodian_id:
-                    maker_custodian_id, size: fill_size, price});
-                // If order on book completely filled:
+                let fill_event = FillEvent{ // Create fill event.
+                    market_id, size: fill_size, price, maker_side: side,
+                    maker_market_order_id: market_order_id, maker,
+                    maker_custodian_id: custodian_id, taker_market_order_id:
+                    ((order_book_ref_mut.counter as u128) << SHIFT_COUNTER)};
+                // If should defer events, push pack event onto queue:
```
elliottdehn, line 2963: typo? push "pack" -> "back"

src/move/econia/sources/market.move
```diff
@@ -3057,16 +3256,18 @@ module econia::market {
                     quote_withdraw, underwriter_id);
             // Declare return assignment variable.
             let self_match_cancel;
+            // Declare empty fill event queue inside of an option.
+            let fill_event_queue_option = option::some(vector[]);
             // Match against order book, storing optionally modified
             // asset inputs, base and quote trade amounts, quote fees
             // paid, and if a self match requires canceling the rest of
-            // the order.
+            // the order. (Increments order book counter).
```
elliottdehn, line 3264: why?

src/move/econia/sources/market.move
```diff
@@ -3044,7 +3241,9 @@ module econia::market {
             base_available, base_ceiling, quote_available, quote_ceiling);
         // Assume no assets traded as a taker.
         let (base_traded, quote_traded, fees) = (0, 0, 0);
-        if (crosses_spread) { // If order price crosses spread:
+        // Evaluate any potential fills across the spread, which may
+        // result in deferred fill events:
```
elliottdehn, line 3245: I'm confused why we have this queue instead of always emitting the event directly to the handle?

src/move/econia/sources/market.move
alnoki: Previously, the order book counter was only incremented when a limit order resulted in a post to the book

Per this PR, the order book counter is also incremented whenever the matching engine is called so that fill events can be assigned a unique order ID

src/move/econia/sources/market.move
alnoki: When a limit order takes and then makes, the assigned order ID needs to encode a pointer into AVL queue memory that is only assigned upon insertion to the order book (the pointer, or "AVL queue access key" is used to lookup on order size change or order cancels).

But, when a limit order takes then makes, the take portion (in matching engine) mutates the AVL queue and so the AVL queue access key is not known until the make portion that posts. Hence, to enforce a common order ID, the events are deferred until after the match operation. See the changes to `place_limit_order()`

alnoki:
The most recent commit provides the following design, which compiles but which hasn't been tested (nor have docs been updated accordingly):

`MarketEvents` under Econia resource account, with lookup via `MarketAccountInfo`:

https://github.com/econia-labs/econia/blob/c5aa3175bd2598674bd46ef0a2b1247194c75369/src/move/econia/sources/market.move#L638-L655

`MatchEvent` design:

https://github.com/econia-labs/econia/blob/c5aa3175bd2598674bd46ef0a2b1247194c75369/src/move/econia/sources/market.move#L565-L576

`MatchEvent` queue behavior inside `match()`:

https://github.com/econia-labs/econia/blob/c5aa3175bd2598674bd46ef0a2b1247194c75369/src/move/econia/sources/market.move#L2989-L3001

`MatchEvent` emission inside `place_limit_order()` after a common order ID has been assigned for both taker and maker sides of order:

https://github.com/econia-labs/econia/blob/c5aa3175bd2598674bd46ef0a2b1247194c75369/src/move/econia/sources/market.move#L3341-L3395

`emit_match_events()` for emitting to market, maker, and taker handles for each `MatchEvent`:

https://github.com/econia-labs/econia/blob/c5aa3175bd2598674bd46ef0a2b1247194c75369/src/move/econia/sources/market.move#L2528-L2603

src/move/econia/sources/market.move
```diff
@@ -628,6 +635,57 @@ module econia::market {
...
+    /// Events are emitted to handle for the market, and to a handle for
+    /// affected user/custodian ID tuple (affected signer in the case of
+    /// a swap, which is is not affiliated with a market account).
+    struct MarketEventHandles has key {
+        /// match() (events deferred to calling function)
+        match_events_by_market:
+            Table<u64, EventHandle<MatchEvent>>,
+        match_events_by_market_account:
+            Table<MarketAccountInfo, EventHandle<MatchEvent>>,
+        /*
```
bogberry, line 656: Event handles for each account should live on the user's account, not the resource account. See Aptos's implementation of [`coin.move`](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-framework/sources/coin.move) for what this should look like.

src/move/econia/sources/market.move
```diff
@@ -552,30 +562,27 @@ module econia::market {
 
     // Structs >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
 
-    /// Emitted when a maker order is placed, cancelled, evicted, or its
-    /// size is manually changed.
-    struct MakerEvent has drop, store {
-        /// Market ID of corresponding market.
+    struct MatchEvent has copy, drop, store {
```
bogberry, line 565: Consider emitting information about fees charged here too?

src/move/econia/sources/market.move
```diff
@@ -2766,13 +2986,19 @@ module econia::market {
...
                         quote_coins, fill_size * lot_size,
                         ticks_filled * tick_size);
-                // Get taker events handle.
-                let taker_handle = &mut order_book_ref_mut.taker_events;
-                // Emit corresponding taker event.
-                event::emit_event(taker_handle, TakerEvent{
-                    market_id, side, market_order_id, maker, custodian_id:
-                    maker_custodian_id, size: fill_size, price});
-                // If order on book completely filled:
+                vector::push_back(event_queue_ref_mut, MatchEvent{
```
bogberry, line 2989: Add space before bracket

bogberry (changes requested):

bogberry:
Also what's the motivation for changing the name from `FillEvent` to `MatchEvent`?

alnoki:
Key implementations in the latest batch of revisions:

New event structs in `user.move`:

https://github.com/econia-labs/econia/blob/308c76f75dfbecb110b250504aa9e843e028de1d/src/move/econia/sources/user.move#L412-L446

Since `market.move` has `user.move` as a dependency, `user.move` offers the following public friend functions:

https://github.com/econia-labs/econia/blob/308c76f75dfbecb110b250504aa9e843e028de1d/src/move/econia/sources/user.move#L1511-L1611

Hence, at the end of `market::place_limit_order()`:

https://github.com/econia-labs/econia/blob/308c76f75dfbecb110b250504aa9e843e028de1d/src/move/econia/sources/market.move#L3310-L3342

Here, `FillEvent`s are emitted inside of the `MarketEventHandles` struct under the Econia resource account, then for each user associated with a fill. After fills are emitted, a `LimitOrderEvent` is similarly emitted at the market- and user-level.

src/move/econia/sources/market.move
alnoki: See https://github.com/econia-labs/econia/pull/321#issuecomment-1622722806 for implementation per your recommendation.

src/move/econia/sources/market.move
alnoki: See `LimitOrderEvent` per https://github.com/econia-labs/econia/pull/321#issuecomment-1622722806

src/move/econia/sources/market.move
alnoki: I agree that the style you are suggesting is more Rust-like and should be used on other codebases moving forward, but per https://github.com/econia-labs/econia#move the Econia Move code is formatted per PEP8, so I am leaning toward consistency within the Econia v4 codebase.

alnoki: See #323

src/move/econia/sources/market.move
```diff
@@ -613,11 +603,11 @@ module econia::market {
...
         bids: AVLqueue<Order>,
-        /// Cumulative number of maker orders placed on book.
+        /// Cumulative number of orders placed.
         counter: u64,
-        /// Event handle for maker events.
+        /// Deprecated field retained for backwards compatibility.
         maker_events: EventHandle<MakerEvent>,
-        /// Event handle for taker events.
+        /// Deprecated field retained for backwards compatibility.
         taker_events: EventHandle<TakerEvent>
```
bogberry, line 611: This doesn't really offer backwards compatibility. We should stop saying this, for reasons I've written [here](https://github.com/econia-labs/econia/pull/315#discussion_r1254266215).

src/move/econia/sources/user.move
```diff
@@ -409,6 +409,42 @@ module econia::user {
...
+        size: u64,
+        price: u64,
+        restriction: u8,
+        self_match_behavior: u8,
+        base_traded_as_taker: u64,
+        quote_traded_as_taker: u64,
+        fees_as_taker: u64,
+        size_posted_as_maker: u64,
+        order_id: u128
+    }
```
bogberry, line 446: Is the idea that this event will be emitted only when a limit order is placed? Or will cancellations and modifications be variations of this? Either way, the name should be changed to make its role more clear.

src/move/econia/sources/user.move
```diff
@@ -1438,6 +1508,108 @@ module econia::user {
...
+                limit_order_events_ref_mut, market_account_id);
+            if (has_handle) {
+                let limit_order_event_handle_ref_mut = table::borrow_mut(
+                    limit_order_events_ref_mut, market_account_id);
+                event::emit_event(limit_order_event_handle_ref_mut, event)
+            }
+        }
+    }
+
+    public(friend) fun get_fill_event(
```
bogberry, line 1540: Name should not begin with `get`, as this is not a getter. `create`?

bogberry (changes requested):

src/move/econia/sources/market.move
alnoki: 69f05ea modifies commentary per `Move.toml` nomenclature:

```toml
[package]
name = "Econia"
version = "4.1.0"             # Per SemVer.
upgrade_policy = "compatible"
```

src/move/econia/sources/user.move
alnoki: 58a7aa9 renames to `PlaceLimitOrderEvent`

src/move/econia/sources/user.move
alnoki: 58a7aa9 renames to `pack_fill_event()`

alnoki:
Key implementations in the latest batch of revisions:

`FillEvent`, `PlaceLimitOrderEvent`, and `PlaceMarketOrderEvent` structs in `user.move`, emitted to a `user::MarketEventHandles` for all of a user's market accounts:

https://github.com/econia-labs/econia/blob/9985609b8f9e8c6a7edac864c9ff1d17eb0ab95d/src/move/econia/sources/user.move#L412-L463

These events are similarly emitted to a `market::MarketEventHandles` at the market level:

https://github.com/econia-labs/econia/blob/9985609b8f9e8c6a7edac864c9ff1d17eb0ab95d/src/move/econia/sources/market.move#L621-L641

Also emitted at the market level are `SwapEvent`s, but since swaps do not take place through a `user::MarketAccount`, swapper-specific event handles are stored under each signing user's account in a `market::SwapperEventHandles`:

https://github.com/econia-labs/econia/blob/9985609b8f9e8c6a7edac864c9ff1d17eb0ab95d/src/move/econia/sources/market.move#L643-L664

Since swaps are possible without a signature, events are only emitted to a user-level `SwapperEventHandles` for a `swap_between_coinstores()`:

https://github.com/econia-labs/econia/blob/9985609b8f9e8c6a7edac864c9ff1d17eb0ab95d/src/move/econia/sources/market.move#L1820-L1851

Note here that the event emitted at the swapper level is returned by the `swap()` function, which emits it to the market-level `MarketEventHandles` regardless of which swap variant is called:

https://github.com/econia-labs/econia/blob/9985609b8f9e8c6a7edac864c9ff1d17eb0ab95d/src/move/econia/sources/market.move#L4048-L4102

src/move/econia/sources/user.move
```diff
@@ -409,6 +409,59 @@ module econia::user {
...
+        size: u64,
+        price: u64,
+        restriction: u8,
+        self_match_behavior: u8,
+        base_traded_as_taker: u64,
+        quote_traded_as_taker: u64,
+        quote_fee_paid_as_taker: u64,
+        size_posted_as_maker: u64,
+        order_id: u128
+    }
```
bogberry, line 449: The initial fill upon placement should not be considered part of the placement—the order is placed, and then immediately filled if it crosses the book. `base_traded_as_taker`, `quote_traded_as_taker`, and `quote_fee_paid_as_taker` can all be part of a fill event that is emitted after the place limit order event. The nuances are important here.

src/move/econia/sources/user.move
```diff
@@ -409,6 +409,59 @@ module econia::user {
...
+        size: u64,
+        price: u64,
+        maker_side: bool,
+        maker: address,
+        maker_custodian_id: u64,
+        maker_order_id: u128,
+        taker: address,
+        taker_custodian_id: u64,
+        taker_order_id: u128
+    }
```
bogberry, line 432: Consider including information about fees here? 

(I said this once [here](https://github.com/econia-labs/econia/pull/321#discussion_r1252595147), but I don't think you ever addressed it.)

src/move/econia/sources/user.move
```diff
@@ -1438,6 +1526,155 @@ module econia::user {
...
+
+    public(friend) fun set_fill_event_taker_order_id(
+        fill_event_ref_mut: &mut FillEvent,
+        taker_order_id: u128
+    ) {
+        fill_event_ref_mut.taker_order_id = taker_order_id
+    }
+
+    public(friend) fun pack_place_limit_order_event(
```
bogberry, line 1616: Why pack?

bogberry (changes requested):

bogberry:
Left some comments, although as I commented in the previous iteration, I need to see how you handle cancellations, modifications, etc. to get the full picture.

alnoki:
# Key structs

Key structs in `user.move` after latest batch of revisions:

https://github.com/econia-labs/econia/blob/88cd68686ccc4232e10850a897bfb856e5a7c12b/src/move/econia/sources/user.move#L421-L470

* Each `FillEvent` is assigned a sequence number for the trade in which it took place (serial counter implemented in `market::match()`
* Fee information for limit orders, market orders, and swaps is broken out into a `TakerFillSummaryEvent` which details the taker address, fee information, amount traded, number of fills for the trade, etc.

Key structs in `market.move`:

https://github.com/econia-labs/econia/blob/88cd68686ccc4232e10850a897bfb856e5a7c12b/src/move/econia/sources/market.move#L621-L663

# Current functionality

* For a limit order that crosses the spread then posts, a `PlaceLimitOrderEvent` and `FillEvent`(s) are emitted at both the user and market level
* For a market order, a `PlaceMarketOrderEvent` and a `FillEvent`(s) at both user and market level
* For a swap order, a `PlaceSwapOrderEvent` and a `FillEvent`(s) at the market level. If the swap is invoked by a signer, then the same events are also emitted at the signer level per the `market::SwapperEventHandles`

# Planned functionality

* A `user::TakerFillSummarEvent` additionally emitted in all such scenarios whenever the number of associated fills is nonzero (struct implemented, but emission not yet implemented)
* A `user::ChangeOrderSizeEvent` struct that is emitted whenever a user or custodian calls `market::change_order_size()`
* A `user::CancelOrderEvent` struct that is emitted whenever a user or custodian calls `market::cancel_order()`, an order is evicted, or an order that has already posted to the book is cancelled due to self match behavior (an enum can declare which of the event types takes place)

src/move/econia/sources/user.move
alnoki: Fees broken out into proposed `TakerFillSummaryEvent` per https://github.com/econia-labs/econia/pull/321#issuecomment-1626162868

src/move/econia/sources/user.move
alnoki: After your previous comment about fees I incorporated the fee information into associated `PlaceLimitOrderEvent`, `PlaceMarketOrderEvent`, or `PlaceSwapOrderEvent`, but per https://github.com/econia-labs/econia/pull/321#issuecomment-1626162868 I have incorporated into the proposed `TakerFillSummaryEvent` in an effort to isolate fee information.

Since the matching engine computes fees using integer division, fees are assessed after all fills for a trade have been executed, to amortize rounding errors. Hence there is no explicit fee associated with any one fill. Fees are also only assessed on the taker side of a trade, so I'm proposing the `TakerFeeSummaryEvent`: this provides information about the amount of quote and base assets traded away/received by the taker, the amount paid in fees, and the number of associated fills. Hence I also added a sequence number to each fill within a trade.

src/move/econia/sources/user.move
alnoki: https://move-language.github.io/move/structs-and-resources.html?highlight=pack#creating-structs

src/move/econia/sources/user.move
alnoki: Per discussion with @qdrs for a concrete example of fills:

Two fills of equal size constitute a trade such that the matching engine assesses 201 indivisible quote subunits of fees. If fees were reported for each `FillEvent`, integer truncation would result in an output 100 subunits of fees per, even though the matching engine assesses 201. Hence a 1 subunit accounting error. This is resolved by the summary fee information in a `TakerFillSummaryEvent`

src/move/econia/sources/user.move
bogberry: No one says that. Why not create?

alnoki:
Key structs in `user.move` after latest batch of revisions:

https://github.com/econia-labs/econia/blob/d341906d51e78dea69455963815f60e63303abde/src/move/econia/sources/user.move#L412-L482

Here, a `ChangeOrderSizeEvent` is emitted whenever a user or a custodian invokes an order size change on an order that has posted to the book (`market::change_order_size_user()` or `market::change_order_size_custodian()`).

A `CancelPostedOrderEvent` is emitted whenever an order that has already posted to the book is cancelled:
* A user or signing custodian manually cancels the order
* The order is evicted
* The order is cancelled due to self-match behavior as either `CANCEL_MAKER` or `CANCEL_BOTH` during execution of a market order or a limit order that fills across the spread

Key structs in `market.move`: 

https://github.com/econia-labs/econia/blob/d341906d51e78dea69455963815f60e63303abde/src/move/econia/sources/market.move#L616-L629

Now only `PlaceSwapOrderEvent`s are indexed at the market-level, since this is the only event that is not guaranteed emission at the user level (`user::MarketEventHandles`).

src/move/econia/sources/user.move
```diff
@@ -409,6 +409,78 @@ module econia::user {
...
+    }
+
+    struct CancelPostedOrderEvent has copy, drop, store {
+        market_id: u64,
+        order_id: u128,
+        user: address,
+        custodian_id: u64,
+        side: bool,
+        is_eviction: bool,
+        is_self_match_cancel: bool
```
bogberry, line 432: It's more common to simply have a `reason` field. See for example
<https://docs.cloud.coinbase.com/exchange/docs/websocket-channels#done>

This would be more in line with industry standards and offers greater flexibility and extensibility.

src/move/econia/sources/user.move
```diff
@@ -409,6 +409,78 @@ module econia::user {
...
+        change_order_size_events:
+            Tablist<u128, EventHandle<ChangeOrderSizeEvent>>,
+        fill_events: Tablist<u128, EventHandle<FillEvent>>,
+        place_limit_order_events:
+            Tablist<u128, EventHandle<PlaceLimitOrderEvent>>,
+        place_market_order_events:
+            Tablist<u128, EventHandle<PlaceMarketOrderEvent>>
+    }
+
+    struct CancelPostedOrderEvent has copy, drop, store {
```
bogberry, line 425: As per @qdrs, I don't think `Posted` here adds anything

alnoki: `Posted` to make it explicit that this event is emitted only when a posted order is cancelled, representing a mutation to order book state.

This is in contrast, for example, to

* A swap that eats through all price levels before fully filling, such that the matching engine halts.
* A limit order that fills multiple times across the spread and then self-matches with the `CANCEL_TAKER` flag, such that the matching engine halts and then the limit order does not post. 

Here the result of each order is fully captured by the combination of `FillEvent`s and, respectively, the `PlaceSwapOrderEvent` or `PlaceLimitOrderEvent` (which includes a `size_posted` field).

Since the logic would be more complex and the events would be redundant if we were to also to emit cancel events for these kinds of cases, I chose to clarify that only posted order cancellations emit events.

Your thoughts?

src/move/econia/sources/market.move
```diff
@@ -628,6 +613,34 @@ module econia::market {
         map: Tablist<u64, OrderBook>
     }
 
+    struct MarketEventHandles has key {
+        /// Not emitted at user level when swapper doesn't sign.
+        place_swap_order_events:
+            Tablist<u64, EventHandle<PlaceSwapOrderEvent>>
```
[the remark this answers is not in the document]
alnoki: The market IDs are not necessarily sequential since a market that was initialized before an on-chain upgrade will initialize an event handle on-the-fly.

However the major drawback I see to switching away from a table-based approach is that storage creations/writes are fixed-cost up to a critical number of bytes (2k at the moment I believe). Aptos Labs discourages vectors for "large" datasets, because the cost to operate on the vector increases linearly for each additional entry.

I believe each `EventHandle` is 48 bytes, so after 42 markets or so get initialized the cost to emit an event grows linearly for every additional market registered. This situation introduces a DoS attack, which is emitted by a Table-based approach.

The rationale for using a `Tablist` instead of a `Table` is to enable iteration over all handles using the Aptos REST API.

Your thoughts?

alnoki:
@bogberry I addressed two of your comments above, standing by for responses. (`CancelPostedOrderEvent` struct renaming is pending discussion.)

b227eb3 adds `reason` field per below:

https://github.com/econia-labs/econia/blob/b227eb35bbd8c4c350336f29bdd72cc02d65b8e5/src/move/econia/sources/user.move#L425-L432

Associated enum:

https://github.com/econia-labs/econia/blob/b227eb35bbd8c4c350336f29bdd72cc02d65b8e5/src/move/econia/sources/user.move#L555-L557

src/move/econia/sources/user.move
alnoki: See https://github.com/econia-labs/econia/pull/321#issuecomment-1632592184

src/move/econia/sources/user.move
bogberry: Then how does the user who posted that limit order know that the order stopped filling because it reached another one of their orders?

The user needs to be able to determine the current status of their orders based on the events they receive through these event handles. The combination of these events do not offer enough information to definitively determine order status.

src/move/econia/sources/market.move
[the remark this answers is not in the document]
bogberry: > The market IDs are not necessarily sequential since a market that was initialized before an on-chain upgrade will initialize an event handle on-the-fly.

Explain? The market IDs are set here, right. How would this ever produce non-sequential IDs?

https://github.com/econia-labs/econia/blob/b227eb35bbd8c4c350336f29bdd72cc02d65b8e5/src/move/econia/sources/registry.move#L1593-L1594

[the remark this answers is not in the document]
alnoki: > > The market IDs are not necessarily sequential since a market that was initialized before an on-chain upgrade will initialize an event handle on-the-fly.
> 
> Explain? The market IDs are set here, right. How would this ever produce non-sequential IDs?
> 
> https://github.com/econia-labs/econia/blob/b227eb35bbd8c4c350336f29bdd72cc02d65b8e5/src/move/econia/sources/registry.move#L1593-L1594

@bogberry per https://github.com/econia-labs/econia/pull/321#discussion_r1261206551

> The market IDs are not necessarily sequential since a market that was initialized before an on-chain upgrade will initialize an event handle on-the-fly.

The market IDs are sequential but the order of initialization of event handles isn't necessarily sequential. E.g. markets 1-5 as legacy markets, the event handle init sequence could look like: 6,7,8,1,2,3,4,5,9,10,...

src/move/econia/sources/user.move
alnoki: > Then how does the user who posted that limit order know that the order stopped filling because it reached another one of their orders?

Also, just to clarify, I use the term "post" to denote an order that is on the book waiting to be filled.

I am confused by your above statement, as I would expect the word "submit" where you used the word "post": if an order stops filling due to a self-match I wouldn't expect it to post

https://www.independentreserve.com/blog/knowledge-base/what-is-a-post-only-order-type

src/move/econia/sources/market.move
```diff
@@ -628,6 +613,34 @@ module econia::market {
...
+    }
+
+    /// Stored under a signing user's account (not market account),
```
elliottdehn, line 622: Shouldn't this be in the `user` module if it's stored under a user's account?

alnoki: It is stored as a resource under the same address as a user's `aptos_framework::account::Account`, which is different from a `econia::user::MarketAccount`: swaps take place outside of an Econia-style `MarketAccount`.

Hence in contrast to an operation through a market account, which requires either a signature or custodian approval, swaps only sometimes require a signature (see https://github.com/econia-labs/econia/pull/321#issuecomment-1624493623)

src/move/econia/sources/user.move
bogberry: Yes, I should have used "placed" instead of "posted" there. Apologies for the imprecise language.

> Presently, the user knows how much posted to the order book via the size_posted field of PlaceLimitOrderEvent, which is not necessarily the same as the size field.

That's not the case I'm asking about, I'm asking about limit orders that never make it to the prior book because they encounter a self match prior to doing so.

I think the partially maker, partially taker case is fine now. Although I do maintain that we should just have `remaining_size` / `amount_filled` separate from `size` like every other exchange on the planet, instead of the `size_posted` field here.

bogberry: > I believe you may be looking for something like the following:
> 
> - Cancel reasons CANCEL_REASON_SELF_MATCH_TAKER and CANCEL_REASON_SELF_MATCH_MAKER corresponding to cancellation of the taker side and/or the maker side depending on self-match behavior
> - Cancel reason CANCEL_REASON_BOOK_EMPTY for when the matching engine halts due to no more possible fills
> - Cancel reason CANCEL_REASON_IMMEDIATE_OR_CANCEL for an IOC order that does not fill across the spread

Where would events with these values be emitted? Under the current event schema, cancel events are only emitted for maker events. This doesn't provide users with enough information to determine the status of their orders via events, as I noted above.

src/move/econia/sources/market.move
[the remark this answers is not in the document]
bogberry: Oh, by "on-chain upgrade" you mean the changes in this PR.

In what situation would this matter though? If markets 1-5 were created before this change, and markets 6-8 were initialized prior to event handles being created for 1-5, when would you want to traverse these event handles in the order 6,7,8,1,2,3,4,5,9,10,...?

alnoki:
Key updates:

Handles stored in `Table` instead of `Tablist`:

https://github.com/econia-labs/econia/blob/ddbdeb4215d80e180e957ba34a103609eb6450e6/src/move/econia/sources/user.move#L412-L422

https://github.com/econia-labs/econia/blob/ddbdeb4215d80e180e957ba34a103609eb6450e6/src/move/econia/sources/market.move#L617-L631

`PlaceLimitOrderEvent` updated with `remaining_size` field:

https://github.com/econia-labs/econia/blob/ddbdeb4215d80e180e957ba34a103609eb6450e6/src/move/econia/sources/user.move#L456-L468

`CancelPostedOrderEvent` renamed to `CancelOrderEvent`:

https://github.com/econia-labs/econia/blob/ddbdeb4215d80e180e957ba34a103609eb6450e6/src/move/econia/sources/user.move#L424-L430

Comprehensive cancel reason enum:

https://github.com/econia-labs/econia/blob/ddbdeb4215d80e180e957ba34a103609eb6450e6/src/move/econia/sources/user.move#L553-L560

Market order/swap order `CANCEL_REASON_...`s

| Situation | `CANCEL_REASON_...` |
|-|-|
| Self match, when behavior indicates taker cancel | `SELF_MATCH_TAKER` |
| Order eats up all liquidity but hasn't fully filled | `NOT_ENOUGH_LIQUIDITY` |
| Continuing to fill would overflow/underflow quote holdings | `MAX_QUOTE_TRADED` |

Limit order `CANCEL_REASON_...`s

| Situation | `CANCEL_REASON_...` |
|-|-|
| Self match, when behavior indicates taker cancel | `SELF_MATCH_TAKER` |
| Remaining size after matching is nonzero for IOC order | `IMMEDIATE_OR_CANCEL` |
| Order still crosses spread after matching engine halts | `MAX_QUOTE_TRADED` |
| Remaining size after matching is less than market `min_size` | `TOO_SMALL_AFTER_MATCHING` |

`PlaceLimitOrderEvent.remaining_size` is emitted as the amount that posts to the book (0 in the case that a cancel event is emitted per above)

src/move/econia/sources/user.move
alnoki: > I think the partially maker, partially taker case is fine now. Although I do maintain that we should just have `remaining_size` / `amount_filled` separate from `size` like every other exchange on the planet, instead of the `size_posted` field here.

@bogberry `PlaceLimitOrderEvent.size_posted` changed to `remaining_size` per https://github.com/econia-labs/econia/pull/321#issuecomment-1635034144

> Where would events with these values be emitted? Under the current event schema, cancel events are only emitted for maker events. This doesn't provide users with enough information to determine the status of their orders via events, as I noted above.

Also per https://github.com/econia-labs/econia/pull/321#issuecomment-1635034144 , `CancelPostedOrderEvent` updated to `CancelOrderEvent` with relevant cancellation reason emission.

src/move/econia/sources/market.move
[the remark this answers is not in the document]
alnoki: @bogberry per https://github.com/econia-labs/econia/pull/321#issuecomment-1635034144 `Tablist` updated to `Table`

src/move/econia/sources/market.move
alnoki: Resolving per #323 merge

alnoki:
Recent optimizations consolidate all event handles to a single global storage resource, per market account, at the user level:

https://github.com/econia-labs/econia/blob/8567b819d4c4ca594f4ad192261af7ed5ed78368/src/move/econia/sources/user.move#L413-L433

At the market level and signing swapper level, events are consolidated per market ID: 

https://github.com/econia-labs/econia/blob/8567b819d4c4ca594f4ad192261af7ed5ed78368/src/move/econia/sources/market.move#L618-L654

These refactors, along with the creation number structs, enable streamlined view functions for getting event handle metadata:

https://github.com/econia-labs/econia/blob/8567b819d4c4ca594f4ad192261af7ed5ed78368/src/move/econia/sources/user.move#L574-L608

https://github.com/econia-labs/econia/blob/8567b819d4c4ca594f4ad192261af7ed5ed78368/src/move/econia/sources/market.move#L874-L927

Hence by looking up the event handle info via a view function accessor (as in #344), market events can be queried by event handle creation number: https://fullnode.devnet.aptoslabs.com/v1/spec#/operations/get_events_by_creation_number

alnoki:
d241602 verified as an on-chain upgrade against [v4.0.2-audited](https://github.com/econia-labs/econia/releases/tag/v4.0.2-audited) in a 1-of-1 v2 multisig at https://explorer.aptoslabs.com/account/0x85738691ac10116f9a516d3bd5c120941f733bbf302fc0a6f6fdd857e938b7c0?network=testnet

qdrs (approved):

bogberry (approved):

PR #330, opened 2023-07-10
Create view function accessor

elliottdehn:
This function requires that the arguments are bcs-encoded and it has a return value in JSON format. It's meant to be wrapped by helper functions and allow granular access to all of the view functions in case we can't get to all of them.

### DIFF
--- a/src/python/sdk/examples/view.py
+++ b/src/python/sdk/examples/view.py
@@ -1,16 +1,208 @@
 from aptos_sdk.account_address import AccountAddress
-from econia_sdk.lib import EconiaViewer;
-from econia_sdk.view.incentives import *;
-from econia_sdk.view.market import *;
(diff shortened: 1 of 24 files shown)


src/python/sdk/examples/view.py
```diff
@@ -0,0 +1,16 @@
+from aptos_sdk.account_address import AccountAddress
+from econia_sdk.lib import EconiaViewer;
+from econia_sdk.view.incentives import *;
+from econia_sdk.view.market import *;
+from econia_sdk.view.registry import *;
+from econia_sdk.view.resource_account import *;
+from econia_sdk.view.user import *
```
bogberry, line 7: <https://docs.python.org/3/tutorial/modules.html#importing-from-a-package>

alnoki:
Closing in lieu of #344

PR #334, opened 2023-07-11
Add reverse market ID lookup view funcs

alnoki:


src/move/econia/sources/registry.move
```diff
@@ -514,6 +519,62 @@ module econia::registry {
...
+    /// * `test_set_remove_check_recognized_markets()`
+    fun get_market_id_base_coin<BaseType, QuoteType>(
+        lot_size: u64,
+        tick_size: u64,
+        min_size: u64,
+    ): Option<u64>
+    acquires Registry {
+        get_market_id(MarketInfo{
+            base_type: type_info::type_of<BaseType>(),
+            base_name_generic: string::utf8(b""),
```
elliottdehn, line 540: Can you help me understand what this string-represented value means?
Is it something like `0xbeef::test_eth::TestETH` where this value is just the string representation of a generic type, or something else?

elliottdehn: From @alnoki:

> The base asset for a market doesn’t have to be a coin
https://econia.dev/overview/registry#underwriters

elliottdehn (approved):

elliottdehn:
image

PR #342, opened 2023-07-14
[Experimental] Try out `@thalalabs/surf`

bogberry:
Trying out Surf.

Surf is a type-safe TypeScript interface & React Hook library for Aptos, created by Thala Labs.
<https://github.com/ThalaLabs/surf>


alnoki:
Nice, it looks like this is decently plug-and-play

PR #347, opened 2023-07-19
Remove cancel event for limit order when all size fills

alnoki:
# Summary of changes

Previously, if a `NO_RESTRICTION` limit order fully filled across the spread, it would emit a `CancelOrderEvent` with `CANCEL_REASON_TOO_SMALL_AFTER_MATCHING`.

Now, if a `NO_RESTRICTION` limit order fills across the spread it will not emit a `CancelOrderEvent` for a full fill. If it takes all liquidity but the remaining size is not big enough to post, it will emit a `CancelOrderEvent` with reason `CANCEL_REASON_TOO_SMALL_AFTER_MATCHING`.

# Versus market orders and swaps

Now, cancel event behavior is consistent across limit orders, market orders, and swap orders. If an order fully fills during the transaction in which it was placed, it will emit A `Place<Limit | Market | Swap>OrderEvent` and `FillEvent`(s), but not a `CancelOrderEvent`.

# Testing

Move source code has been tested to 100% coverage, but there is no way to validate event fields inside a Move unit test. I documented Aptos' event testing deficiency in October (https://github.com/aptos-labs/aptos-core/issues/5369), and in May I cited insufficient event testing as the likely reason for an error in Aptos' own code that I corrected through a PR (https://github.com/aptos-labs/aptos-core/pull/8448).

The inability to write Move unit tests that assert event emissions is a contributing factor to this PR's changes not getting slated into #321 , and seeing as this issue was uncovered during discussions about projected event emission schedules, it is suggested that all event emission cases and field be asserted during end-to-end tests in Python and/or Rust.

### DIFF
--- a/src/move/econia/sources/market.move
+++ b/src/move/econia/sources/market.move
@@ -890,9 +890,6 @@ module econia::market {
     /// Order cancelled because it was on the taker side of an fill
     /// where self match behavior indicated cancelling the taker order.
     const CANCEL_REASON_SELF_MATCH_TAKER: u8 = 7;
-    /// Order cancelled because after matching across the spread the
-    /// remaining order size was too small for the market.
-    const CANCEL_REASON_TOO_SMALL_AFTER_MATCHING: u8 = 8;
     /// Flag to cancel taker order only during a self match.
     const CANCEL_TAKER: u8 = 3;
     /// Critical tree height above which evictions may take place.
@@ -3248,7 +3245,7 @@ module econia::market {
     /// * `test_place_limit_order_crosses_ask_self_match_cancel()`
     /// * `test_place_limit_order_crosses_bid_exact()`
     /// * `test_place_limit_order_crosses_bid_partial()`
-    /// * `test_place_limit_order_crosses_bid_partial_cancel()`
+    /// * `test_place_limit_order_crosses_bid_partial_post_under_min()`
     /// * `test_place_limit_order_evict()`
     /// * `test_place_limit_order_no_cross_ask_user()`
     /// * `test_place_limit_order_no_cross_ask_user_ioc()`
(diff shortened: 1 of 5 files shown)


src/move/econia/sources/market.move
```diff
@@ -3451,11 +3451,10 @@ module econia::market {
...
                              CANCEL_REASON_MAX_QUOTE_TRADED);
-            } else {
-                if (remaining_size < order_book_ref_mut.min_size) {
-                    option::fill(&mut cancel_reason_option,
-                                 CANCEL_REASON_TOO_SMALL_AFTER_MATCHING);
-                }
+            } else if ((remaining_size > 0) &&
+                       (remaining_size < order_book_ref_mut.min_size)) {
+                option::fill(&mut cancel_reason_option,
+                             CANCEL_REASON_TOO_SMALL_AFTER_MATCHING);
```
bogberry, line 3457: I think `min_size` should be enforced only when the order is first placed. If an order of a size greater than `min_size` is placed, and it fills in such a way that after the initial fill the remaining size falls below this threshold, that's not an issue.

src/move/econia/sources/market.move
```diff
@@ -3470,7 +3469,7 @@ module econia::market {
         let market_order_id =
             ((order_book_ref_mut.counter as u128) << SHIFT_COUNTER);
         // If order eligible to post:
-        if (option::is_none(&cancel_reason_option)) {
+        if (option::is_none(&cancel_reason_option) && (remaining_size > 0)) {
```
bogberry, line 3472: This combined with 

https://github.com/econia-labs/econia/blob/f5d5c133063bbe88735766e7800cffcaf95b6654/src/move/econia/sources/market.move#L3507-L3508

mean that any order that is cancelled here will have `remaining_size` set to zero. However, this shouldn't be the case. If, for instance, an IOC order partially filled and then cancelled due to the entire order not being immediately fillable, the remaining size should still not be zero as the order was not fully filled.

bogberry (changes requested):

src/move/econia/sources/market.move
alnoki: @bogberry addressed as of f2c1777

src/move/econia/sources/market.move
alnoki: @bogberry addressed as of f2c1777

bogberry: Ok, but in the future please link the commit in which the fix was made so I can look at the diff

bogberry (approved):

src/move/econia/sources/market.move
alnoki: @bogberry will do, thanks for clarifying

PR #350, opened 2023-07-20
ECO-227: Hide depth chart when viewing on short screen

bogberry:
Also fixes the layout shift that used to occur while the site was waiting for TradingView to load.


chiumax (commented):
LGTM, interesting use of variants.

chiumax (approved):

PR #355, opened 2023-07-21
ECO-234 ECO-233 ECO-232 minor card title and depth chart padding fixes

chiumax:
image


chiumax:
okay i merged from main -- i shouldn't have broken anything

PR #356, opened 2023-07-21
ECO-235 consistent limit market order card sizing

chiumax:


bogberry:
Hmmmm the empty space here looks a bit weird. Might be better to have 
- The empty space above the "Est. Fee" section so that there's minimal layout shift between the limit/market tabs? or
- Assign minimum height to the parent div of the order form + trade history so that together they don't fall below a certain height, but the trade history table still grows in height when switching to the market order tab?

@xorkiki what do you think?

Screenshot 2023-07-24 at 15 15 39

xorkiki:
Hmm.. I am leaning toward the second option. I just sent over a video in the frontend slack channel for reference here!

I think it looks like a waste of space to have empty space there when the trade history could easily occupy it. 

Thoughts? @bogberry @BriungRi

bogberry:
I'm leaning towards the second option too. Just pushed some changes that I think should be what you're looking for, please take a look!

bogberry (approved):

PR #366, opened 2023-07-24
[ECO-247] Add event emission unit tests, with uncovered cancel reasons

alnoki:
# Background

#321 and several associated follow up PRs added comprehensive event emissions for market events like `PlaceLimitOrderEvent`. At the time, however, event emissions could not verified using the Move unit testing framework, and several minor issues were uncovered through discussions/manual testing:

* #347
* #360

Aptos Labs recently enabled unit testing of events in Move via https://github.com/aptos-labs/aptos-core/pull/9181, and as such this PR updates the Move unit testing suite to test event emissions.

# Methodology

Before this PR, Move code was already tested to 100% coverage, and as such this PR simply adds event assert statements to the existing coverage suite where applicable. Notably, this includes all non-expected-failure tests listed under the doc comments for:

* `market::match()`
* `market::place_limit_order()`
* `market::place_market_order()`,
* `market::swap_between_coinstores()`
* `market::swap_coins()`
* `market::swap_generic()`

Auxiliary event scenarios are also validated for manual order size changes and manual order cancellations, as well as registry events.

# Gaps addressed

The two issues listed under [background](#background) are now tested for and would have been quickly caught if event testing had been available at the time of submission of #321. Two additional edge cases were discovered during testing for this PR, both of which are cancel reasons that only apply to swaps. All applicable cancel reasons for all scenarios (limit order, market order, swap, manual cancel, eviction, self match maker and taker cancel) are now tested. The cancel reason logic and associated documentation is included in this PR per:

* `CANCEL_REASON_TOO_SMALL_TO_FILL_LOT`: https://github.com/econia-labs/econia/pull/366/commits/dd968e97400c4bb41ad556aca35d59f90dfa3973
* `CANCEL_REASON_VIOLATED_LIMIT_PRICE`: https://github.com/econia-labs/econia/pull/366/commits/a894fdb35be649a8a4e2b5dbc2cb3b6faf85c95f

### DIFF
--- a/src/move/econia/sources/market.move
+++ b/src/move/econia/sources/market.move
@@ -3743,11 +3743,12 @@ module econia::market {
     ///
     /// # Expected value testing
     ///
-    /// * `test_place_market_order_max_base_adjust_buy_user()` TODO
-    /// * `test_place_market_order_max_base_buy_user()` TODO
-    /// * `test_place_market_order_max_base_sell_custodian()` TODO
-    /// * `test_place_market_order_max_quote_buy_custodian()` TODO
-    /// * `test_place_market_order_max_quote_sell_user()` TODO
+    /// * `test_place_market_order_max_base_below_buy_user()`
+    /// * `test_place_market_order_max_base_buy_user()`
+    /// * `test_place_market_order_max_base_sell_custodian()`
+    /// * `test_place_market_order_max_quote_buy_custodian()`
+    /// * `test_place_market_order_max_quote_sell_user()`
+    /// * `test_place_market_order_remove_event_handles()`
     ///
     /// # Failure testing
     ///
@@ -8200,7 +8201,6 @@ module econia::market {
                 place_market_order_user<BC, QC>(
             &user, market_id, integrator, direction_taker, size_taker,
             self_match_behavior_taker);
-        // Assert event streams.
         let taker_order_id = order_id_no_post(3);
         // Assert events.
         assert!(user::get_place_limit_order_events_test(
@@ -11578,7 +11578,7 @@ module econia::market {
     /// Verify state updates, returns for market buy when user specifies
     /// base trade amount that is less than max possible, under
     /// authority of signing user.
-    fun test_place_market_order_max_base_adjust_buy_user()
+    fun test_place_market_order_max_base_below_buy_user()
     acquires OrderBooks {
         // Initialize markets, users, and an integrator.
         let (user_0, user_1) = init_markets_users_integrator_test();
@@ -11592,10 +11592,10 @@ module econia::market {
         let size_post           = 10; // Maker order size.
         let size_match          = MIN_SIZE_COIN; // Taker order size.
         let base_match          = size_match * LOT_SIZE_COIN;
-        let base_post           = size_post  * LOT_SIZE_COIN;
+        let base_post           = size_post * LOT_SIZE_COIN;
         let price               = integrator_divisor * taker_divisor;
         let quote_match         = size_match * price * TICK_SIZE_COIN;
-        let quote_post          = size_post  * price * TICK_SIZE_COIN;
+        let quote_post          = size_post * price * TICK_SIZE_COIN;
         let integrator_share    = quote_match / integrator_divisor;
         let econia_share        =
             quote_match / taker_divisor - integrator_share;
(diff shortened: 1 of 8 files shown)


src/move/econia/sources/market.move
```diff
@@ -5007,9 +5168,28 @@ module econia::market {
             price, restriction, self_match_behavior, &custodian_capability);
         // Check if order is tail of corresponding price level.
         assert!(is_local_tail_test(market_id, side, market_order_id), 0);
+        // Assert event streams.
+        assert!(user::get_cancel_order_events_test(
```
elliottdehn, line 5172: Why assert the cancel events here if we're only changing the order size? Seems kind of random. Do orders self-cancel as a result of a size change to a size too low? Is that tested, and either way I feel like trying to set an order to a size too low should just throw exception instead of canceling.

Docs don't mention any cancel-by-size-change: https://econia.dev/off-chain/events#changing-order-size

src/move/econia/sources/market.move
```diff
@@ -5150,8 +5330,27 @@ module econia::market {
             restriction, self_match_behavior);
         // Check if order is tail of corresponding price level.
         assert!(is_local_tail_test(market_id, side, market_order_id), 0);
+        // Assert event streams.
+        assert!(user::get_cancel_order_events_test(
```
elliottdehn, line 5334: Here too with the confusing cancel asserts.

src/move/econia/sources/user.move
```diff
@@ -3271,11 +3300,132 @@ module econia::user {
...
+    acquires MarketEventHandles {
+        let market_event_handles_map_ref =
+            &borrow_global<MarketEventHandles>(user).map;
+        let market_account_id = get_market_account_id(market_id, custodian_id);
+        table::borrow(market_event_handles_map_ref, market_account_id)
+    }
+
+    #[test_only]
+    /// Return a `ChangeOrderSizeEvent` with the indicated fields.
+    public fun create_change_order_size_event_test(
```
elliottdehn, line 3319: Why create these helpers? Having them makes it seem like there's something more interesting going on than... just building the struct. Don't see why these (create_change_order_size_event_test, create_place_limit_order_event_test, create_place_market_order_event_test) are here.

src/move/econia/sources/market.move
```diff
@@ -6086,6 +6373,68 @@ module econia::market {
...
+                    taker_order_id,
+                    fee,
+                    0
+                )
+            ], 0);
+        assert!(user::get_cancel_order_events_test(
+            market_id, maker_address, NO_CUSTODIAN) == vector[], 0);
+        assert!(get_place_swap_order_events_market_test(
+            market_id) == vector[
+                PlaceSwapOrderEvent{
```
elliottdehn, line 6415: Why no helper for these other events? (advocating for consistency here)

src/move/econia/sources/market.move
```diff
@@ -6390,6 +6834,86 @@ module econia::market {
                        quote_coins);
         // Assert order book counter.
         assert!(get_order_book_counter(market_id) == 3, 0);
+        let taker_order_id = order_id_no_post(3);
+        // Assert events.
+        assert!(user::get_place_limit_order_events_test(
```
elliottdehn, line 6839: TBH I would just have a helper for asserting all events given the market id, account address (optional?), and custodian id being asserted. Then you would just pass your expected vectors in and it would assert all of the event types. This way we know for a fact that we've asserted all events that get emitted and all events that don't get emitted--by using such a helper.

elliottdehn (changes requested):
See note about adding a helper to unify the assertions being made at each location, perhaps have it accept a structure that makes it very clear what's being asserted. I'd call it `assert_events` or something.

src/move/econia/sources/market.move
alnoki: Cancel order events are asserted as empty here because user-specific size changes are handled internal to user memory as a cancel followed by an immediate placement

This is the rationale for having `user::CANCEL_REASON_SIZE_CHANGE_INTERNAL`, and is related to #360

Asserting there are no cancel events is done to enforce that a cancel event isn't erroneously emitted due to the user-side logic

src/move/econia/sources/market.move
alnoki: Same rationale as for https://github.com/econia-labs/econia/pull/366#discussion_r1277962396

elliottdehn:
I would also advise you to write out above the assertion what's being expected, i.e:

```
- Expect Cancel event emitted to maker because reasons
- Expect no other events emitted to taker or maker.
```

It's hard to review this in GitHub because you have to click several times to see what the test name is in order to figure out what should be asserted and then check that this is correct. Just a note though.

src/move/econia/sources/user.move
alnoki: These helpers are here so that the structs can be created from a call inside the `market` module

Move doesn't allow structs to be created from outside the module in which they are defined, so the `user` module needs to have public APIs that the market module can use

The events are emitted at the user level, but need to be tested in market-level scenarios (hence nearly all tests for this PR are in `market` module)

src/move/econia/sources/market.move
alnoki: Unlike the structs from https://github.com/econia-labs/econia/pull/366#discussion_r1277970860, the `PlaceSwapOrderEvent` is only ever emitted to an event handle that is define in the market module

Hence it can be created directly here using curly brace packing syntax

src/move/econia/sources/market.move
alnoki: If a similar logical flow were taking place during runtime I would agree with the recommendation for abstraction

But since these are tests I lean toward using the least sophisticated methods possible e.g. keeping the code dumb and procedural

This way the abstracted test helpers don't themselves also need to be tested

alnoki:
> I would also advise you to write out above the assertion what's being expected, i.e:
> 
> ```
> - Expect Cancel event emitted to maker resulting from X.
> - Expect no other events emitted to taker or maker resulting from Y, Z.
> ```
> 
> It's hard to review this in GitHub because you have to click several times to see what the test name is in order to figure out what SHOULD be asserted and then check that this is correct. Just a note though. I'll have to pull it down to review it anyway.

This should be much easier to review in VS Code using the pull requests extension https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github

That way you can open the same editor side by side and review the function badge name

src/move/econia/sources/market.move
```diff
@@ -11994,7 +15554,7 @@ module econia::market {
         let min_base            = 0;
         let max_base            = 0;
         let min_quote           = 0;
-        let max_quote           = quote_trade + TICK_SIZE_COIN;
+        let max_quote           = quote_trade * TICK_SIZE_COIN;
```
elliottdehn, line 15557: I kind of wish, instead of declaring all of these at the top of the test every time and copy pasting the block, that this were instead a structure returned by a default value method (or several, such as for bid/ask). Return the structure, and then any modifications made are clear and explicit. So basically you have known defaults and explicit deviations from those defaults. Right now, it's difficult to _understand_ this section in each test. What's different and interesting? Default struct + explicit mutations = much nicer.

elliottdehn (approved):
I feel comfortable approving this after reviewing most (75%) of the changes closely because for something to be wrong in the specific case of an assert statement two things would have to be true:
- You're asserting something as true that shouldn't be.
- The code is doing the wrong thing and satisfying the assertion.

I wasn't able to find any missing cases or incorrect assertions.

src/move/econia/sources/market.move
alnoki: Good point on the default vs explicit mutations

I was avoiding abstractions to try and keep the test code simple to follow, e.g. so that we don't have to test the test code

But agreed that this style introduces some capacity for replicated errors

On this PR I don't think it's a worthwhile exercise to go back and abstract things

But on future testing suites there is certainly room for improvement

src/move/econia/sources/market.move
```diff
@@ -4196,10 +4216,22 @@ module econia::market {
             limit_price,
             order_id: market_order_id
         };
+        // Check if violated limit price.
+        let violated_limit_price = if (!liquidity_gone) {
```
bogberry, line 4220: Why check if the limit price is too high or low for the order to be filled after attempting to match? Why not before?

bogberry (changes requested):

src/move/econia/sources/market.move
alnoki: @bogberry `violated_limit_price` is only checked after attempted matching because orders that match, incur slippage, and then cancel due to a limit price violation only violate the limit price after matching, not before. Forr this case, if the check were performed before attempted matching then the violated limit price cancel reason would be missed.

Orders that violate the limit price before attempted matching result in `match()` halting without matching, then trip the `violated_limit_price` check here.

So by checking after attempted matching the result is the same for both cases, a cancel event with `CANCEL_REASON_VIOLATED_LIMIT_PRICE`

bogberry: This strikes me as a very convoluted way of performing this check. During the matching process you know when you’ve reached a price level that falls above or below the limit price, and there is still remaining size on the orderbook to fill. 

Attempting to fill the order and mutating the orderbook, borrowing the orderbook _again_ to get the top of the orderbook you were at a moment ago to check against the limit price of the order adds unnecessary steps to perform what should be a very simple check.

Alternatively you could check against the orderbook without performing any mutations first to see if the entire order is fillable. This way you can have just a `borrow_global` at the initial step and abort without ever calling `borrow_global_mut` if the order is impossible to fill.

alnoki: > During the matching process you know when you’ve reached a price level that falls above or below the limit price, and there is still remaining size on the orderbook to fill

@bogberry indeed, see 6a1deeb

bogberry: Why is there not a similar cancel reason emitted for fill-or-kill limit orders? Seems like there's a check missing here.

alnoki: There are no "fill-or-kill" orders but there is a `FILL_OR_ABORT` restriction:

Unlike an immediate-or-cancel order, for example, which emits a `PlaceLimitOrderEvent`, optionally `FillEvent`(s), and optionally a `CancelOrderEvent`, if a fill-or-abort order does not fill completely across the spread the entire transaction aborts because of the complexity associated with rolling back fills. Hence “abort”

A CEX has a cheap database and can rollback memory and simulate fills etc., but for the matching engine it would be difficult to do this and cost more gas

Hence it tries filling and if at the end it doesn’t meet the condition it blows up the transaction with an abort, which means that in terms of events emitted, state changes, it’s like the transaction was never even placed

bogberry (approved):

PR #371, opened 2023-07-26
Use CSS linear gradient for orderbook bars

bogberry:
Depends on #370.

### DIFF
--- a/src/typescript/frontend/src/components/OrderbookTable.tsx
+++ b/src/typescript/frontend/src/components/OrderbookTable.tsx
@@ -40,12 +40,24 @@ const Row: React.FC<{
     baseCoinDecimals: BigNumber(marketData.base?.decimals || 0),
   });
 
+  const barPercentage = (level.size * 100) / highestSize;
+  const barColor =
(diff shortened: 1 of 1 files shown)


src/typescript/frontend/src/components/OrderbookTable.tsx
```diff
@@ -40,13 +40,25 @@ const Row: React.FC<{
...
         setPrice(price.toString());
       }}
+      style={{
+        background: `linear-gradient(
+          to left,
+          ${barColor},
+          ${barColor} ${barPercentage}%,
+          transparent ${barPercentage}%
+        )`,
+      }}
```
chiumax, line 61: jw is this better practice and setting widths manually like i did before an antipattern?
regardless everything looks great

chiumax (approved):

src/typescript/frontend/src/components/OrderbookTable.tsx
chiumax: oh NVM didn't see ur comment in slack. diregard!

PR #374, opened 2023-07-26
[ECO-193] Update minor things in Python SDK for pip push

elliottdehn:
Pip doesn't allow an empty readme so let's put something of most use and least maintenance there.
Need to update the project fields because @qdrs is no longer an author.
This is "release v1" as far as I know.
link to page: https://pypi.org/project/econia-sdk/

### DIFF
--- a/src/python/sdk/pyproject.toml
+++ b/src/python/sdk/pyproject.toml
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "econia-sdk"
-version = "0.1.0"
+version = "1.0.0"
 description = ""
-authors = ["qdrs <qdrs@proton.me>"]
+authors = ["egd <elliott@econialabs.com>"]
 readme = "README.md"
 packages = [{ include = "econia_sdk" }]
 
--- a/src/python/sdk/README.md
+++ b/src/python/sdk/README.md
@@ -1,0 +1,1 @@
+See [documentation](https://econia.dev/off-chain/python-sdk).


src/python/sdk/pyproject.toml
```diff
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "econia-sdk"
-version = "0.1.0"
+version = "1.0.0"
 description = ""
-authors = ["qdrs <qdrs@proton.me>"]
+authors = ["egd <elliott@econialabs.com>"]
```
alnoki, line 5: Please include "alnoki <43892045+alnoki@users.noreply.github.com>"

@bogberry @CRBl69 what to include for you?

alnoki (changes requested):

src/python/sdk/pyproject.toml
qdrs: image

alnoki: just discussed offline this is for point of contact

we are going to use developers@econialabs.com

alnoki: sorry for not clarifying about this earlier
i was really just thinking about moving forward how to make sure the right people get the right emails
oversight on my part

alnoki (approved):

src/python/sdk/pyproject.toml
alnoki: See https://github.com/econia-labs/econia/pull/380

alnoki (approved):

PR #375, opened 2023-07-26
[ECO-195] Update Pdocs for Python SDK: `pdoc -o docs econia_sdk`

elliottdehn:
Here I've just run the command in-title to generate the docsite for the Python SDK. Will add docstrings to all of the view functions at a later point in time.


alnoki (changes requested):
Can you update the README for the Python SDK too?

I believe this will get included on the PyPI

elliottdehn:
> Can you update the README for the Python SDK too?
> 
> I believe this will get included on the PyPI

I just updated the README on this branch to what it is in #374 but it looks like this tool does not pick up the README.

alnoki (approved):

bogberry:
You can add the README to pdoc by including 

```
"""
.. include:: ../README.md
"""
```

to the top-level `__init__.py` as documented here:

<https://pdoc.dev/docs/pdoc.html>

PR #378, opened 2023-07-27
ECO-133: Add sorting to orders table, overflow when order rows do not fit

bogberry:
Adds
- Sorting by column when clicking header of orders table
- Overflow on y-axis when there are too many order rows to fit inside the table height

Notes
- It seems like there are no easy solutions for adding a border on the bottom of the table header that does not scroll with the body? I ended up adding a pseudoelement on the `<tr>` element inside `<thead>`, but open to suggestions for better solutions
- Need some icons for sorting direction indicators. I've put in some unicode characters as placeholders, but this shouldn't be the final version
- Should order statuses other than "open" be color coded? Figma only shows this variant.


chiumax:
maybe we can opt to use chevron down and up from heroicons? (we're already using them)

chiumax:
hmm, also the sorting icon shifts elements over

bogberry:
Cool, added some margin but I think it looks good!

chiumax (approved):
sweet, thanks for getting this done

PR #381, opened 2023-07-27
Use box-shadow for sticky table header

BriungRi:
Possible improvements:

1. Use tailwind -- done
2. Access color variable instead of hardcode -- done


src/typescript/frontend/src/scratch.ts.bak
chiumax, line 1: this file here on purpose?

chiumax:
uses tailwind: https://tailwindcss.com/docs/box-shadow#arbitrary-values

chiumax (approved):
anyways, i like this

bogberry:
I think this is better, nice work!

PR #384, opened 2023-07-27
ECO-144: frontend toast when tx fail

chiumax:
did some snooping their codebase and didn't find any references for the option `checkSuccess` (though it does show up in intellisense as a type).
Regardless, they had an example for error handling that I ported over
https://github.com/aptos-labs/aptos-wallet-adapter/tree/main/packages/wallet-adapter-react#signandsubmittransactionpayload


bogberry (approved):
LGTM!

PR #386, opened 2023-07-28
[ECO-283] Change decimals and name of faucet test currency

elliottdehn:
tETH has 18 decimals and thus you can only have 18 whole units of tETH before overflowing the u64. Let's use a different currency (tAPT/TestAPT) with 8 decimals so that this lines up better. Everyone who's run this code has run into this issue.

It might be kind of confusing to have tAPT and APT adjacent to one another, but I think it's ok.


bogberry (changes requested):
Why not just use testnet APT?

elliottdehn:
> Why not just use testnet APT?

We're limited in how many APT we can request to mint, so it's much easier to be able to request lots fake APT than 1 APT at a time from a "real" faucet.

alnoki (changes requested):
As just discussed during a team call, let's switch to `eAPT` ("Econia APT") and `eUSDC` ("Econia USDC") to avoid confusion that might arise from the `t` prefix (e.g. Tortuga liquid staking derivative)

src/python/sdk/examples/trade.py
```diff
@@ -127,8 +127,8 @@ def get_aptos_faucet_url() -> str:
     get_econia_address()
 )  # See https://econia.dev/ for up-to-date per-chain addresses
 FAUCET_ADDR = get_faucet_address()  # See (and deploy): /econia/src/move/faucet
-COIN_TYPE_ETH = f"{FAUCET_ADDR}::test_eth::TestETH"
-COIN_TYPE_USDC = f"{FAUCET_ADDR}::test_usdc::TestUSDC"
+COIN_TYPE_EAPT = f"{FAUCET_ADDR}::example_apt::ExampleAPT"
+COIN_TYPE_USDC = f"{FAUCET_ADDR}::example_usdc::ExampleUSDC"
```
alnoki, line 131: For consistency how about `COIN_TYPE_EUSDC`?

alnoki (changes requested):

alnoki (changes requested):

alnoki (changes requested):

src/move/faucet/sources/example_apt.move
```diff
@@ -1,16 +1,16 @@
...
     use aptos_std::string;
     use econia_faucet::faucet;
 
-    struct TestETH {}
+    struct ExampleAPT {}
 
-    const NAME: vector<u8> = b"Test Ether";
-    const SYMBOL: vector<u8> = b"tETH";
-    const DECIMALS: u8 = 18;
+    const NAME: vector<u8> = b"Example Aptos coin";
```
alnoki, line 7: "Example Aptos Coin" (capital C)

alnoki (changes requested):

alnoki (approved):

bogberry (approved):
Yup all good

PR #387, opened 2023-07-28
[ECO-280] Create cspell dictionary and add words to dictionary

elliottdehn:
Both Aptos and Econia appear often in our texts, but are detected as error words by cspell CLI. Let's fix that.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -1,0 +1,12 @@
+{
+    "version": "0.2",
+    "ignorePaths": [],
+    "dictionaryDefinitions": [],
+    "dictionaries": [],
+    "words": [
+        "aptos",
+        "econia"
+    ],
+    "ignoreWords": [],
+    "import": []
+}


qdrs:
What an utter waste of time? Are company man hours being spent on stuff like this? Launch the product first

bogberry (changes requested):
There seems to be a similiar list in `.vscode/settings.json`, so one of these should be removed in favor of the other.

bogberry:
But I don't think we should be running spellcheck on all of our code. Using names that aren't in the English language in code is pretty common, and I don't want to be adding words to a dictionary every time I do that

alnoki:
This PR touches on some recent DevOps discussions:

* Checks should ultimately be done in CI to reduce manual overhead (e.g. someone has to remember to run a check like a spellchecker)
* I do think that having a spellchecker is useful though, as it raises the degree of professionalism (Aptos' docs use a dictionary file and perform spellchecking on PRs)

bogberry:
We absolutely should not be running spellcheck in CI and rejecting PRs that have spelling errors. Using character strings that are not English words in code is EXTREMELY common, and we shouldn't be expected to add every single instance of a character string that isn't an English word into a dictionary to get a PR approved is ridiculous.

elliottdehn:
Closing this because it should be done by CI if at all.

PR #389, opened 2023-07-31
New event types

qdrs:


src/rust/db/src/lib.rs
```diff
@@ -19,61 +22,136 @@ pub mod schema;
 
 pub type Result<T> = std::result::Result<T, DbError>;
 
-pub fn establish_connection(url: String) -> Result<PgConnection> {
-    PgConnection::establish(&url).map_err(DbError::ConnectionError)
+pub struct EconiaDb(PgConnection);
```
bogberry, line 25: What value does `EconiaDb` add? What's the issue with just using the `PgConnection` on its own

src/rust/Cargo.toml
```diff
@@ -3,13 +3,24 @@ members = ["aggregator", "api", "db", "types"]
...
-chrono = "^0.4.24"
-dotenvy = "^0.15.6"
-envy = "^0.4.2"
-futures-util = "^0.3.26"
-serde = { version = "^1.0.149", features = ["derive"] }
-serde_json = "^1.0.93"
-sqlx = { version = "^0.6.2", features = ["runtime-tokio-rustls"] }
-thiserror = "^1.0.38"
-tokio = "^1.25.0"
+axum = "0.6.19"
```
bogberry, line 6: Why do dependencies that are only used in one workspace member need to be in the workspace root?

src/rust/db/migrations/2023-03-16-025536_create_markets/down.sql
```diff
@@ -1,15 +1,15 @@
-drop table recognized_market_events;
+DROP TABLE recognized_market_events;
```
bogberry, line 1: Isn't lowercase easier to read than uppercase?

src/rust/db/src/lib.rs
qdrs: Then the methods for adding are implemented on the db type rather than just taking conn over and over as the first parameter. Feels cleaner.

src/rust/Cargo.toml
qdrs: Even if you put a dependency only in 1 crate, its beneficial to have it at the workspace level because the entire workspace shares the same Cargo.lock file, so by doing this it stops someone in the future placing a conflicting version dependency in a member of the workspace.

src/rust/db/migrations/2023-03-16-025536_create_markets/down.sql
qdrs: Recommended formatter did this, and tbh SQL should be upper case as that's the standard.

src/rust/db/src/lib.rs
bogberry: But if you need to use diesel, which you definitely do, you still need to access `PgConnection`, and you're forced to do `&mut *db` in place of `conn`. Overall this is just adding needless abstraction in code where you can't really abstract the database connection away.

src/rust/Cargo.toml
bogberry: Hmmmm sure

src/rust/db/migrations/2023-03-16-025536_create_markets/down.sql
bogberry: What formatter are you using? I have sqlfluff set up, and running it with the config file I have added shouldn't change capitalization

bogberry: SQL is agnostic towards casing so it doesn't really matter, and I would much rather NOT HAVE MY SQL SCREAMING AT ME LIKE THIS

src/rust/db/src/lib.rs
qdrs: done i reverted this bit

src/rust/db/migrations/2023-03-16-025536_create_markets/down.sql
qdrs: done its all lowercase now

alnoki (approved):

bogberry (approved):

PR #393, opened 2023-07-31
ECO-141 ECO-140 ECO-139 ECO-138 ECO-137 - market & limit order validation

chiumax:


bogberry:
I seem to get "invalid tick size" or "invalid lot size" no matter what sizes I enter. Can you check to see if the checks are being conducted properly?

bogberry:
Also, the error messages for these should be displayed similarly to how "required" is, not as toasts. Sorry for not making this explicit in the Linear tasks

bogberry (changes requested):
☝️

bogberry (changes requested):
I get `SyntaxError: Cannot convert 0.1 to a BigInt` when entering an order size of 0.1. I assume you're missing a conversion somewhere?

bogberry:
I think "Invalid size. Lot size for market is X" is good enough for as an error message. I think they should go elsewhere to have lot size and tick size explained

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
```diff
@@ -60,6 +64,76 @@ export const LimitOrderEntry: React.FC<{
     };
     const orderSide = orderSideMap[side];
 
+    // validation
+    const _rawValueSize = toRawCoinAmount(
+      values.size ?? 0,
+      marketData?.base?.decimals ?? 0,
```
bogberry, line 70: Best to check whether you have all the values you need at the top instead of using optional chaining and null coalescing everywhere. Performing validation using 0 if the values here are undefined is kinda pointless

src/typescript/frontend/src/utils/formatter.ts
```diff
@@ -76,3 +76,12 @@ export const averageOrOtherPriceLevel = (
...
 };
+
+export const canBeBigInt = (num: string | number | bigint | boolean) => {
+  try {
+    BigInt(num);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
```
bogberry, line 87: Not really a fan of using try catch in this way - I think if you are going to use a try catch block you should just do it in the original function

But in this case, I think `Number.isInteger(Number(num))` does what you need?

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
```diff
@@ -60,6 +64,76 @@ export const LimitOrderEntry: React.FC<{
...
+    // limit buy -- make sure user has enough quote balance
+    if (orderSide === "bid") {
+      const isValid =
+        BigInt(
+          toRawCoinAmount(
+            rawQuoteBalance.toString(),
+            marketData?.base?.decimals ?? 0,
+          ),
+        ) >=
+        BigInt(rawValueSize) * BigInt(rawValuePrice); // how to get price? esp if market order
```
bogberry, line 121: I think you end up with too many decimal places here

If e.g. base has 8 decimals and the quote has 6 decimals, `rawValueSize` is `size * 10^8`, `rawValuePrice` is `price * 10^6`, so I believe you have `size * price * 10^14` here?

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
```diff
@@ -60,6 +64,76 @@ export const LimitOrderEntry: React.FC<{
...
+      if (!isValid) {
+        setError("price", { message: "INSUFFICIENT QUOTE BALANCE" });
```
bogberry, line 123: I prefer showing insufficient balance errors on size

bogberry (changes requested):
I've left comments on individual lines, but overall
- There's a lot of unnecessary conversions. If you're gonna use `BigDecimal` at the beginning, might as well stick with it and not convert to `BigInt`s along the way until the very end
- There's not really a point of coalescing missing values to 0 here, since the results you get from using 0 instead isn't really relevant. Better to throw an error if the application is in a state it shouldn't be in
- Try to exit the function as early as possible and avoid doing calculations you might not end up needing

src/typescript/frontend/src/utils/formatter.ts
chiumax: Sure, I can move it -- `Number.isInteger(Number(100000000000000000000000000.1))` evaluates to `true`

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
chiumax: yep, and i also have `quote *10^14` so it balances out

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
chiumax: just change to say `INSUFFICIENT BALANCE`?

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
chiumax: good point

chiumax:
> I've left comments on individual lines, but overall
> 
> * There's a lot of unnecessary conversions. If you're gonna use `BigDecimal` at the beginning, might as well stick with it and not convert to `BigInt`s along the way until the very end
> * There's not really a point of coalescing missing values to 0 here, since the results you get from using 0 instead isn't really relevant. Better to throw an error if the application is in a state it shouldn't be in
> * Try to exit the function as early as possible and avoid doing calculations you might not end up needing

I agree, which is why I wanted to use BigNumber initially

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
bogberry: Ohhhh sorry, missed that this uses `base.decimals`, nvm

src/typescript/frontend/src/components/trade/OrderEntry/LimitOrderEntry.tsx
bogberry: yeah I guess so

src/typescript/frontend/src/utils/formatter.ts
bogberry: oh lol

bogberry:
Yeah I was more thinking not ever using `BigNumber` and doing all of the calculations as `BigInt`s. You'd have to use a regex in the beginning while size and price are still strings to make sure you were getting a valid int so it's an extra step but I think it's more performant

But there's no point if you're converting to `BigNumber` in `toRawCoinAmount` and converting back

bogberry (approved):

PR #394, opened 2023-07-31
[ECO-271] Add comments and use black to format

elliottdehn:
I've copied over the comments from Move code and put them into the Python SDK. I've skipped adding links to the move documentation since that's going to be either duplicate information or irrelevant information to a Python SDK user. `black` has picked up changes in some files because we have dueling formatters I believe.

It's not verbatim copies of the comments -- I've gone and made minor edits and exclusions where doing so makes sense, such as excluding references to making the function private in Move for XYZ reason.


alnoki (changes requested):
Will you please update the PR name to include [ECO-271] ?

alnoki (changes requested):
One more thing:

Will you please update the headers to be consistent with how they are in the entry package?

https://github.com/econia-labs/econia/blob/b48e4bd645fe5fcc746ff2abe21935b36e33e06c/src/python/sdk/econia_sdk/entry/incentives.py#L27-L28

E.g. use

```
The header:
* Point 1
* Point 2
```

instead of 

```
# The header
* Point 1
* Point 2
```

elliottdehn:
> One more thing:
> 
> Will you please update the headers to be consistent with how they are in the entry package?
> 
> https://github.com/econia-labs/econia/blob/b48e4bd645fe5fcc746ff2abe21935b36e33e06c/src/python/sdk/econia_sdk/entry/incentives.py#L27-L28
> 
> E.g. use
> 
> ```
> The header:
> * Point 1
> * Point 2
> ```
> 
> instead of
> 
> ```
> # The header
> * Point 1
> * Point 2
> ```

Done

elliottdehn:
> Will you please update the PR name to include [ECO-271] ?

done

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -140,6 +218,18 @@ def get_open_orders(
     n_asks_max: int = _HI_64,
     n_bids_max: int = _HI_64,
 ) -> dict:
+    """
+    Index order book for given market ID into "asks" and "bids"
+    vectors. Vectors sorted by price-time priority.
+
+    Parameters
```
alnoki, line 225: typo: `Parameters:` per https://github.com/econia-labs/econia/pull/394#pullrequestreview-1555321875

alnoki: And similarly for all new docstrings

PR #402, opened 2023-08-01
[ECO-279] Develop decimal guidance

elliottdehn:
Multiple times clients and ourselves have stumbled on the lots and ticks abstraction (along with min size) and struggled to come up with the values that achieve the outcomes we want. I've added some utilities to help with this along with a guide.


elliottdehn:
closed as duplicate #403

PR #404, opened 2023-08-02
General housekeeping for Python SDK

bogberry:
One thing to consider: black defaults to 88 character lines.

They lay out their reasoning here, and I would favor sticking to it, but are 80 character lines important enough to anyone here that they feel it is worth overriding this setting?

<https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length>

### DIFF
--- a/src/python/sdk/.gitignore
+++ b/src/python/sdk/.gitignore
@@ -1,18 +1,42 @@
-qlite3
-db.sqlite3-journal
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
 
-# Flask stuff:
-ipfile.lock
+# Pycharm user-specific stuff
+.idea/**/workspace.xml
+.idea/**/tasks.xml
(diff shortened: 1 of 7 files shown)


src/python/sdk/.gitignore
```diff
@@ -1 +1,83 @@
...
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
```
elliottdehn, line 66: did you get this file from somewhere?

elliottdehn (commented):
LGTM, thanks!

alnoki (approved):
@bogberry thanks for this

FYI the original `scripts.sh` Python housekeeping I set up was 80 character lines simply to be compliant with PEP8, but I see no problem with sticking to black's defaults

Approving, but deferring merge per oustanding comments from @elliottdehn

PR #406, opened 2023-08-02
[ECO-279] Develop decimal guidance

elliottdehn:
Multiple people have stumbled across lot size, tick size, and min size and rightly become confused. I've developed here helpers for the Python SDK (as well as documentation) that calculate lot/tick size and min size for the user. I wanted something to help with price without the chance of it being misused or truncation ruining someone's day so I provided a helper to determine the minimum quote per lot, that is price granularity.

### DIFF
--- a/doc/doc-site/docs/off-chain/python-sdk.md
+++ b/doc/doc-site/docs/off-chain/python-sdk.md
@@ -56,14 +56,13 @@ However, each value in each enum is associated with a constant that exists in th
 
 ## `econia_sdk.utils.decimals`
 
-This package contains a few helpers for calculating market creation parameters (lot size, tick size, and min size).
-The intent is to allow one to express their desired sizes in decimal notation, and have that converted to integer notation.
-Nota bene that these expect reasonable inputs.
-For example: converting 0.000001 unit size for 3 decimals returns 1 subunit size, which is 0.001 unit size, instead of an error.
+This package contains a few helpers for calculating market parameters (lot size, tick size, and min size).
+The intent is to allow one to express their desired sizes in decimal notation, and have that converted to integer notation. These helpers make an effort to check for reasonable inputs, perfect results in all cases would be impossible due to integer division and decimal truncation.
 Price conversion utilities are left out for this reason: truncation, even a small percent's worth, would make them too dangerous.
 
 Let's walk through an example of configuring a market using these utilities.
 If you'd like to follow along, install the Econia SDK, run the Python interpreter and import the SDK using...
+
 ```bash
 pip3 install econia-sdk
 python3
@@ -79,39 +78,47 @@ Market configuration (beyond base type and quote type) consists of 3 integer val
 | min size | number of lots | the minimum limit order size |
 
 In order to proceed, we must decide the granularity of base and quote sizes as well as the minimum limit order size for our market.
-One must consider that "price" in the exchange is an integer expressed in terms of "ticks per lot" so the granularity (size) of ticks and lots relates to the prices that can be expressed.
-That is, tick size and lot size beget a constant price granularity.
-In general the desire is to have one _tick_ be "small" relative to one **_lot_** in terms of value, since the minimum expressible price is 1 tick per lot, then 2 ticks and so forth.
-Let's use `eAPT` and `eUSDC` used in the faucet below as example base and quote types respectively for a market.
-We know that eAPT (the base type) has 8 decimals and eUSDC (the quote type) has 6 decimals:
+Consider that "price" in the exchange is an integer expressed in terms of "ticks per lot", which means that tick size and lot size affect the prices that can be expressed.
+That is: price granularity is a function of lot size _and_ tick size.
+In general the desire is to have one _tick_ be "small" relative to one **_lot_** in terms of value, since the minimum expressible price is 1 tick per lot. The second smallest possible price is 2 ticks per lot, and so on.
+Let's use `eAPT` and `eUSDC` used in the faucet below as example base and quote types, respectively, for a market.
+We know that `eAPT` (the base type) has 8 decimals and `eUSDC` (the quote type) has 6 decimals:
+
 ```
 >>> base_decimals = 8
 >>> quote_decimals = 6
 ```
 
 We'd like to be able to express small prices with high granularity.
-Let's try using 0.001 eAPT for the granularity base coin sizes, and 0.01 eUSDC (one penny) for the granularity of quote coin sizes.
+Let's try using 0.001 eAPT for the granularity of base coin sizes, and 0.01 eUSDC (one penny) for the granularity of quote coin sizes.
 Does this work?
 Let's find out!
+
 ```
 >>> get_min_quote_per_base(0.001, 0.01)
 10.0
 ```
+
+Here, we're asking what the smallest representable price for 1 `eAPT` is in terms of `eUSDC`, given the lot size and tick sizes specified above.
+
 There are 1000 lots in 1 `eAPT` (=1/0.001) and the minimum expressible price is 1 tick per lot.
-We know that 1 tick is worth 1 cent.
-Thus, the granularity of price in human terms is \$10 per eAPT--thus the minimum price is \$10/eAPT, followed by \$20/eAPT and so forth.
+Since 1 tick is worth 1 cent, this means that the minimum representable price for 1000 lots (1 `eAPT`) is 1000 ticks, or 10 `eUSDC`.
+The granularity of price in human terms is $10 per `eAPT`.
+That is, the minimum price is \$10/eAPT, the second-lowest expressible price is \$20/eAPT, and so on.
 Given that (real) APT is right now \$7, we can assess that this configuration would not be workable or appropriate!
 
 The problem is fixable by using a more granular tick size, say 0.00001 instead of 0.01.
 Checking our price granularity now shows better results:
+
 ```
 >>> get_min_quote_per_base(0.001, 0.00001)
 0.01
 ```
 
-That's a price granularity of 1 cent, since the result here is in quote units and 1 eUSDC is \$1.
+That's a price granularity of 1 cent, since the result here is in quote units and 1 eUSDC is $1.
 Now that we have the minimum base and quote size decimals we'd like to use, we're ready to configure the market.
 Let's get the lot size and tick size:
+
 ```
 >>> lot_size = get_lot_size(0.001, base_decimals)
 >>> lot_size
@@ -121,18 +128,25 @@ Let's get the lot size and tick size:
 10
 ```
 
-We also need a minimum size. Let's say orders must have a minimum of 1 whole APT in order to post, given the lot size of 100000 from above:
+We also need a minimum size.
+Let's say orders can only be submitted if they are for at least 1 whole `eAPT`, using the lot size of 100000 from above:
+
 ```
 >>> get_min_size(1.0, base_decimals, lot_size)
 1000
 ```
 
-That means we use use a minimum size of 1000 lots to configure our market, so the configuring triple would be:
-- Lot size: 100000 (subunits of base)
-- Tick size: 10 (subunits of quote)
-- Min size: 1000 (lots of base)
+That means we use use a minimum size of 1000 lots to configure our market, so our market parameters would be:
 
-This gives us a market with a price granularity of 1 cent and minimum order size of 1 eAPT! Can you see how one would get a price granularity of 0.1 cents instead?
+| parameter | value | units |
+| -- | -- | -- |
+| Lot size | 100000 | Subunits of base |
+| Tick size | 10 | Subunits of quote |
+| Min size | 1000 | Lots of base |
+
+This gives us a market with a price granularity of 1 cent and minimum order size of 1 eAPT!
+Can you see how one would get a price granularity of 0.1 cents instead?
+
 # Other Contents
 
 ## `examples.trade`
--- a/doc/doc-site/docs/off-chain/python-sdk.md
+++ b/doc/doc-site/docs/off-chain/python-sdk.md
@@ -71,11 +71,11 @@ python3
 
 Market configuration (beyond base type and quote type) consists of 3 integer values: lot size, tick size, and min size.
 
-| value | units | description |
-| -- | -- | -- |
-| lot size | subunits of base type | the granularity of base sizes |
+| value     | units                  | description                    |
+| --------- | ---------------------- | ------------------------------ |
+| lot size  | subunits of base type  | the granularity of base sizes  |
 | tick size | subunits of quote type | the granularity of quote sizes |
-| min size | number of lots | the minimum limit order size |
+| min size  | number of lots         | the minimum limit order size   |
 
 In order to proceed, we must decide the granularity of base and quote sizes as well as the minimum limit order size for our market.
 Consider that "price" in the exchange is an integer expressed in terms of "ticks per lot", which means that tick size and lot size affect the prices that can be expressed.
@@ -104,7 +104,7 @@ Here, we're asking what the smallest representable price for 1 `eAPT` is in term
 There are 1000 lots in 1 `eAPT` (=1/0.001) and the minimum expressible price is 1 tick per lot.
 Since 1 tick is worth 1 cent, this means that the minimum representable price for 1000 lots (1 `eAPT`) is 1000 ticks, or 10 `eUSDC`.
 The granularity of price in human terms is \$10 per `eAPT`.
-That is, the minimum price is \$10/`eAPT`, the second-lowest expressible price is $20/`eAPT`, and so on.
+That is, the minimum price is \$10/`eAPT`, the second-lowest expressible price is \$20/`eAPT`, and so on.
 Given that (real) `APT` is right now \$7, we can assess that this configuration would not be workable or appropriate!
 
 The problem is fixable by using a more granular tick size, say 0.00001 instead of 0.01.
@@ -142,11 +142,11 @@ Let's say orders can only be submitted if they are for at least 1 whole `eAPT`, 
 
 That means we use use a minimum size of 1000 lots to configure our market, so our market parameters would be:
 
-| parameter | value | units |
-| -- | -- | -- |
-| Lot size | 100000 | Subunits of base |
-| Tick size | 10 | Subunits of quote |
-| Min size | 1000 | Lots of base |
+| parameter | value  | units             |
+| --------- | ------ | ----------------- |
+| Lot size  | 100000 | Subunits of base  |
+| Tick size | 10     | Subunits of quote |
+| Min size  | 1000   | Lots of base      |
 
 This gives us a market with a price granularity of 1 cent and minimum order size of 1 `eAPT`!
 Can you see how one would get a price granularity of 0.1 cents instead?
--- a/doc/doc-site/docs/off-chain/python-sdk.md
+++ b/doc/doc-site/docs/off-chain/python-sdk.md
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
 Note that Move doesn't have enum types so unlike most of the above, these do not map directly to Move.
 However, each value in each enum is associated with a constant that exists in the Move code.
 
+## `econia_sdk.utils.decimals`
+
+This package contains a few helpers for calculating market creation parameters (lot size, tick size, and min size).
+The intent is to allow one to express their desired sizes in decimal notation, and have that converted to integer notation.
+Nota bene that these expect reasonable inputs.
+For example: converting 0.000001 unit size for 3 decimals returns 1 subunit size, which is 0.001 unit size, instead of an error.
+Price conversion utilities are left out for this reason: truncation, even a small percent's worth, would make them too dangerous.
+
+Let's walk through an example of configuring a market using these utilities.
+If you'd like to follow along, install the Econia SDK, run the Python interpreter and import the SDK using...
+```bash
+pip3 install econia-sdk
+python3
+>>> from econia_sdk.utils.decimals import *
+```
+
+Market configuration (beyond base type and quote type) consists of 3 integer values: lot size, tick size, and min size.
+
+| value | units | description |
+| -- | -- | -- |
+| lot size | subunits of base type | the granularity of base sizes |
+| tick size | subunits of quote type | the granularity of quote sizes |
+| min size | number of lots | the minimum limit order size |
+
+In order to proceed, we must decide the granularity of base and quote sizes as well as the minimum limit order size for our market.
+One must consider that "price" in the exchange is an integer expressed in terms of "ticks per lot" so the granularity (size) of ticks and lots relates to the prices that can be expressed.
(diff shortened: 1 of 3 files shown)


doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
 Note that Move doesn't have enum types so unlike most of the above, these do not map directly to Move.
 However, each value in each enum is associated with a constant that exists in the Move code.
 
+## `econia_sdk.utils.decimals`
+
+This package contains a few helpers for calculating market creation parameters (lot size, tick size, and min size).
```
alnoki, line 59: How about just "market parameters", since these persist beyond creation?

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+The intent is to allow one to express their desired sizes in decimal notation, and have that converted to integer notation.
+Nota bene that these expect reasonable inputs.
```
alnoki, line 61: "Note that these helpers expect reasonable inputs."

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+For example: converting 0.000001 unit size for 3 decimals returns 1 subunit size, which is 0.001 unit size, instead of an error.
```
alnoki, line 62: This example is unclear to me: are you saying that if you specify 0.000001 for a coin that has 3 decimals, it will get railed to 0.001?

If this is the case then think there *should* be an error

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+Price conversion utilities are left out for this reason: truncation, even a small percent's worth, would make them too dangerous.
+
+Let's walk through an example of configuring a market using these utilities.
+If you'd like to follow along, install the Econia SDK, run the Python interpreter and import the SDK using...
+```bash
+pip3 install econia-sdk
+python3
+>>> from econia_sdk.utils.decimals import *
+```
```
alnoki, line 71: I think it's fine here to have the user follow along with a `pip` install and the interpreter, but there should probably be something here to clarify how in production we use poetry to enforce pinned package dependencies

alnoki: Also `import *` is terrible practice but again making an exception here because it's a walkthrough

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+
+Market configuration (beyond base type and quote type) consists of 3 integer values: lot size, tick size, and min size.
+
+| value | units | description |
+| -- | -- | -- |
+| lot size | subunits of base type | the granularity of base sizes |
+| tick size | subunits of quote type | the granularity of quote sizes |
+| min size | number of lots | the minimum limit order size |
```
alnoki, line 79: How about running `mdformat` on this file to make the table more readable?

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+
+In order to proceed, we must decide the granularity of base and quote sizes as well as the minimum limit order size for our market.
+One must consider that "price" in the exchange is an integer expressed in terms of "ticks per lot" so the granularity (size) of ticks and lots relates to the prices that can be expressed.
```
alnoki, line 82: First consider that "price" in the exchange is an integer expressed in terms of "ticks per lot", which means that tick size and lot size affect the prices that can be expressed.

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+That is, tick size and lot size beget a constant price granularity.
```
alnoki, line 83: That is, price granularity is a function of lot size *and* tick size.

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+In general the desire is to have one _tick_ be "small" relative to one **_lot_** in terms of value, since the minimum expressible price is 1 tick per lot, then 2 ticks and so forth.
```
alnoki, line 84: In general the desire is to have one _tick_ be "small" relative to one **_lot_** in terms of value, since the minimum expressible price is 1 tick per lot.
The second smallest possible price is 2 ticks per lot, and so on.

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+Let's use `eAPT` and `eUSDC` used in the faucet below as example base and quote types respectively for a market.
```
alnoki, line 85: "types, respectively, for"

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+We know that eAPT (the base type) has 8 decimals and eUSDC (the quote type) has 6 decimals:
```
alnoki, line 86: Use monospace `eAPT` and `eUSDC` here for consistency

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+```
+>>> base_decimals = 8
+>>> quote_decimals = 6
+```
+
+We'd like to be able to express small prices with high granularity.
+Let's try using 0.001 eAPT for the granularity base coin sizes, and 0.01 eUSDC (one penny) for the granularity of quote coin sizes.
```
alnoki, line 93: "for the granularity of base coin sizes"

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+Does this work?
+Let's find out!
+```
+>>> get_min_quote_per_base(0.001, 0.01)
+10.0
+```
+There are 1000 lots in 1 `eAPT` (=1/0.001) and the minimum expressible price is 1 tick per lot.
```
alnoki, line 100: It wasn't immediately clear to be what this function does (I'm reading this walkthrough before the python module), so I'd suggest adding something before the current sentence, like:

Here we are essentially asking, "if I select a lot size of 0.001 `eAPT` and a tick size of 0.01 `eUSDC`, what will be the smallest representable price for 1 `eAPT`, in terms of `eUSDC`?"

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+We know that 1 tick is worth 1 cent.
```
alnoki, line 101: Since 1 tick is worth 1 cent, this means that the minimum representable price for 1000 lots (1 `eAPT`) is 1000 ticks, or 10 `eUSDC`.

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+Thus, the granularity of price in human terms is \$10 per eAPT--thus the minimum price is \$10/eAPT, followed by \$20/eAPT and so forth.
```
alnoki, line 102: ", the second-lowest expressible price is $20/`eAPT`, and so on."

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+Given that (real) APT is right now \$7, we can assess that this configuration would not be workable or appropriate!
+
+The problem is fixable by using a more granular tick size, say 0.00001 instead of 0.01.
+Checking our price granularity now shows better results:
+```
+>>> get_min_quote_per_base(0.001, 0.00001)
+0.01
+```
+
+That's a price granularity of 1 cent, since the result here is in quote units and 1 eUSDC is \$1.
```
alnoki, line 112: `eUSDC`

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
...
+      This should be of the base coin, not the quote coin!
+    * `base_coin_decimals`: The number of decimals one whole unit of the
+      base has (e.g. USDC has 6 decimals, ETH has 18).
+    * `lot_size`: The size in subunits of one lot of base coin.
+    """
+    smallest_subunits = smallest_decimal_size * (10 ** base_coin_decimals)
+    return math.ceil(smallest_subunits / lot_size)
+
+
+def get_min_quote_per_base(
```
alnoki, line 51: How about we rename this to `get_min_quote_per_base_nominal`, since here we are dealing with nominal amounts rather than with the integer `Coin.value` representations?

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
+import math
+
+def get_lot_size(smallest_decimal_unit: float, coin_decimals: int) -> int:
```
alnoki, line 3: How about `get_lot_size_integer` to clarify that this is outputting the integer value to use for market registration?

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Parameters:
+    * `smallest_decimal_unit`: Decimal (e.g. 0.001 for one thousandth) of
+      the smallest possible unit relative to a whole unit.
+    * `coin_decimals`: The number of decimals one whole unit of the coin
+      has (e.g. USDC has 6 decimals, ETH has 18).
+    """
+    return math.ceil(10**(coin_decimals+math.log10(smallest_decimal_unit)))
+
+def get_tick_size(smallest_decimal_unit: float, coin_decimals: int) -> int:
```
alnoki, line 16: Same here, I suggest `get_tick_size_integer`

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Parameters:
+    * `smallest_decimal_unit`: Decimal (e.g. 0.001 for one thousandth) of
+      the smallest possible unit relative to a whole unit.
+    * `coin_decimals`: The number of decimals one whole unit of the coin
+      has (e.g. USDC has 6 decimals, ETH has 18).
+    """
+    return get_lot_size(smallest_decimal_unit, coin_decimals)
+
+def get_min_size(
```
alnoki, line 29: `get_min_size_integer` per above

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+```
+>>> lot_size = get_lot_size(0.001, base_decimals)
+>>> lot_size
+100000
+>>> tick_size = get_tick_size(0.00001, quote_decimals)
+>>> tick_size
+10
+```
+
+We also need a minimum size. Let's say orders must have a minimum of 1 whole APT in order to post, given the lot size of 100000 from above:
```
alnoki, line 124: Technically an order can post below the minimum size if it first fills across the spread, so how about

"Let's say orders can only be submitted if they are for at least 1 whole `eAPT`, using the lot szie of `100000` from above:

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+```
+>>> get_min_size(1.0, base_decimals, lot_size)
+1000
+```
+
+That means we use use a minimum size of 1000 lots to configure our market, so the configuring triple would be:
```
alnoki, line 130: "so our market parameters would be:"

doc/doc-site/docs/off-chain/python-sdk.md
alnoki: Also line break after first sentence

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+- Lot size: 100000 (subunits of base)
+- Tick size: 10 (subunits of quote)
+- Min size: 1000 (lots of base)
```
alnoki, line 133: How about a table here?

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
+
+This gives us a market with a price granularity of 1 cent and minimum order size of 1 eAPT! Can you see how one would get a price granularity of 0.1 cents instead?
```
alnoki, line 135: Line break after first sentence

Also, how about "This gives us a market with a price granularity of 0.01 `eUSDC`, a minimum order size of 1 `eAPT`, and an `eAPT` granularity of 0.001 `eAPT`!"

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,85 @@ This package contains various enum types useful for parsing and referring to imp
...
 # Other Contents
```
alnoki, line 136: There should be a newline here, but if you run `mdformat` it should take care of itself

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
+import math
+
+def get_lot_size(smallest_decimal_unit: float, coin_decimals: int) -> int:
+    """
+    Given a decimal representation of the smallest possible unit and the
+    whole unit's decimals (e.g. ETH has 18 decimals), return the lot size.
+
+    Parameters:
+    * `smallest_decimal_unit`: Decimal (e.g. 0.001 for one thousandth) of
```
alnoki, line 9: Python docs themselves list some issues with using floats for base-10 financial applications, so I'd advise using a Python decimal under the hood

Here, have the user pass arguments as a string, e.g. `'0.0001'`

Then inside the function first assert that the input is a string, and convert the string to a decimal.Decimal

This should do away with potential errors

alnoki: Also then do this for all other inputs that are not strictly integers

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,67 @@
...
+      the smallest possible unit relative to a whole unit.
+    * `coin_decimals`: The number of decimals one whole unit of the coin
+      has (e.g. USDC has 6 decimals, ETH has 18).
+    """
+    return math.ceil(10**(coin_decimals+math.log10(smallest_decimal_unit)))
+
+def get_tick_size(smallest_decimal_unit: float, coin_decimals: int) -> int:
+    """
```
alnoki, line 17: Here again I suggest `get_tick_size_integer`, using string input then cast to decimal.Decimal

Also do type checking to enforce that what should be an int is an int

alnoki (changes requested):
Looks good so far, I like the doc style of showing something that is problematic and using it as a teaching moment. A few things:

1. How about running the markdown file through `mdformat` (at the end of course)?
2. We now have official Python formatters per https://github.com/econia-labs/econia/pull/404, so run those too (at the end of course)
3. I left some suggestions on the python module about input, type checking, and name conventions. I think these should be applied in all instances and the interpreter examples re-generated accordingly
4. After the existing example, can you add a smaller one with a case for `wBTC`, which I believe has 9 decimals? Due to its high market price there are some trade-offs in choosing market parameters, as will become clear
5. Can you also do an example of something with parameters that aren't powers of 10, e.g. base granularity 0.5 and price granularity 0.125?

src/python/sdk/econia_sdk/utils/decimals.py
elliottdehn: I think the type checking python does (which I think is none) is fine given there's type hints. We shouldn't be type checking in type-hinted functions.

doc/doc-site/docs/off-chain/python-sdk.md
elliottdehn: I ran it but it didn't format the table.

elliottdehn:
> After the existing example, can you add a smaller one with a case for wBTC, which I believe has 9 decimals? Due to its high market price there are some trade-offs in choosing market parameters, as will become clear
I don't think this necessary, we have extensive documentation on market parameters elsewhere in the docs where more depth like this would be good. This is not the place.
> Can you also do an example of something with parameters that aren't powers of 10, e.g. base granularity 0.5 and price granularity 0.125?
Same as above, I don't think it's interesting enough to be worth bothering the reader with. Just stick to powers of ten in this section and go into more detail elsewhere. I think the reader understands that math works for non-tens.

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,103 @@ This package contains various enum types useful for parsing and referring to imp
 Note that Move doesn't have enum types so unlike most of the above, these do not map directly to Move.
 However, each value in each enum is associated with a constant that exists in the Move code.
 
+## `econia_sdk.utils.decimals`
+
+This package contains a few helpers for calculating market parameters (lot size, tick size, and min size).
+The intent is to allow one to express their desired sizes in decimal notation, and have that converted to integer notation. These helpers make an effort to check for reasonable inputs, perfect results in all cases would be impossible due to integer division and decimal truncation.
```
alnoki, line 60: 1. Line break after first sentence
2. I suggest second sentence become: "Perfect conversions are not always possible due to potential integer division and truncation, so these helpers functions attempt to check for reasonable inputs"

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,103 @@ This package contains various enum types useful for parsing and referring to imp
...
+Price conversion utilities are left out for this reason: truncation, even a small percent's worth, would make them too dangerous.
```
alnoki, line 61: "Price conversion utilities are left out to avoid introducing truncation effects."

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,103 @@ This package contains various enum types useful for parsing and referring to imp
...
+
+```bash
+pip3 install econia-sdk
+python3
+>>> from econia_sdk.utils.decimals import *
+```
+
+Market configuration (beyond base type and quote type) consists of 3 integer values: lot size, tick size, and min size.
+
+| value     | units                  | description                    |
```
alnoki, line 74: For consistency with other tables in the docs, how about "Value", "Units", "Description"?

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,103 @@ This package contains various enum types useful for parsing and referring to imp
...
+| --------- | ---------------------- | ------------------------------ |
+| lot size  | subunits of base type  | the granularity of base sizes  |
```
alnoki, line 76: Capitalize first letter in table entry

doc/doc-site/docs/off-chain/python-sdk.md
```diff
@@ -54,6 +54,103 @@ This package contains various enum types useful for parsing and referring to imp
...
+Let's say orders can only be submitted if they are for at least 1 whole `eAPT`, using the lot size of 100000 from above:
+
+```
+>>> get_min_size_integer("1.0", base_decimals, lot_size)
+1000
+```
+
+That means we use use a minimum size of 1000 lots to configure our market, so our market parameters would be:
+
+| parameter | value  | units             |
```
alnoki, line 145: Same table capitalization standards as above

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,106 @@
...
+    Given a decimal representation of the smallest possible unit and the
+    whole unit's decimals (e.g. ETH has 18 decimals), return the lot size.
+
+    Parameters:
+    * `smallest_decimal_unit`: Decimal (e.g. "0.001" for one thousandth) of
+      the smallest possible unit relative to a whole unit.
+    * `coin_decimals`: The number of decimals one whole unit of the coin
+      has (e.g. USDC has 6 decimals, ETH has 18).
+    """
+    smallest_decimal = Decimal(smallest_decimal_unit)
```
alnoki, line 16: To avoid potential float issues I suggest a type check on all `str` via a helper func like below:

```python
def verify_decimal_input(decimal_input: str):
    if type(decimal_input) is not str:
        raise ValueError("Decimal input should be a string")
```

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,106 @@
...
+    smallest_subunits = 10 ** (coin_decimals + math.log10(smallest_decimal))
```
alnoki, line 17: I also suggest the same for the decimals, e.g. `verify_int_input` because (un?)fortuantely Python isn't strictly typed

src/python/sdk/econia_sdk/utils/decimals.py
```diff
@@ -0,0 +1,106 @@
...
+      the smallest possible size relative to a whole unit. Note that this
+      maybe different (larger) than the smallest possible unit (lot size).
+      This should be of the base coin, not the quote coin!
+    * `base_coin_decimals`: The number of decimals one whole unit of the
+      base has (e.g. USDC has 6 decimals, ETH has 18).
+    * `lot_size`: The size in subunits of one lot of base coin.
+    """
+    smallest_decimal = Decimal(smallest_decimal_size)
+    smallest_subunits = (
+        smallest_decimal * (10**base_coin_decimals)
```
alnoki, line 57: I'm surprised the formatter gave `(10**base_coin_decimals)`...

alnoki: I ran it myself and got the same result, so I guess black removes spaces in exponentiation like this

However there were other changes, so will you please re-run the formatter once more after addressing the above changes?

alnoki (changes requested):
Almost there, just a few minor requests

PR #418, opened 2023-08-08
[ECO-137 - ECO-141] Order validation logic changes

bogberry:
Suggested changes for #393.

- Checks for missing values at the very beginning to avoid usage of optional chaining and null coalescing later on
- Moves return statements to the earliest point they can be called, preventing unnecessary calculations
- Reduces the number of conversions, especially to bigint, necessary to perform these validations
- Fixes number of decimal places for checking quote balance on limit bid orders


chiumax:
this makes sense and is what I wanted to do initially

chiumax (approved):

PR #419, opened 2023-08-08
[ECO-245] Edit orders documentation a bit

elliottdehn:
I've removed the math which is kind of distracting in favor of using the new Python helpers with a link to their implementation for the curious.

The rest of the sections are honestly fine in my opinion. Usage of Python to calculate things is fine, it was the Latex that was distracting.

### DIFF
--- a/doc/doc-site/docs/overview/orders.md
+++ b/doc/doc-site/docs/overview/orders.md
@@ -215,61 +215,58 @@ Econia does not use a critical height of 1. The actual critical height is define
 
 ## Units and market parameters
 
+There are a few important properties one cares about when creating and configuring a market, in rough order of priority:
+
+#### Price granularity
+
+Market participants should be able to express prices that are close enough together to be effectively continuous while being far enough apart to reflect meaningful movement with a single increment.
+
+- **Too granular:** Almost every order occupies its own price level, and user interfaces struggle to display a useful segment of the order book before running out of visual real-estate. Maximum price may be too low (not common).
+- **Too coarse:** A single increment (or decrement) in price reflects a dramatic shift in value whose magnitude exceeds those of smaller movements that are still meaningful to market participants. Minimum price may be too high (common).
+- **Just right:** Prices are effectively continuous in that any meaningful movement is always expressible as some increment or decrement in price. Orders coalesce into useful price levels, and the minimum/maximum prices are not relevant concerns to market action.
+
+#### Minimum order size
+
+Market participants should be able to create a limit order as long as it's for a meaningful amount of asset.
+Each individual limit order affected by a market order adds to the network fee cost of the market order's execution.
+Thus, it's in every participant's interest that every limit order has a meaningful enough size to include in the execution of a market order.
+
+- **Too low:** The limit order books become polluted with small orders and it becomes difficult or impossible to execute a market order without paying a significant network fee. Denial-of-Service is a concern at this point because it becomes possible to "shut down" trading by spamming limit orders.
+- **Too high:** Some or many meaningful market participants are cut out of trading action and thus must take their liquidity somewhere else.
+- **Just right:** Nobody with good intentions is cut out of liquidity provision based on the amount of asset available to them, as long as their amount is meaningful.
+
+#### Order size granularity
+
+Market participants should be able to express any amount as an order size that is close to or equal to the amount it reflects.
+One increment or decrement in order size should be a meaningful but not overwhelming difference in value.
+
+- **Too granular:** Price granularity is too coarse due to lack of ability to express the price of one size increment (or decrement) in terms smaller than a single subunit of asset.
+- **Too coarse:** A single increment (or decrement) in size reflects a dramatic shift in value whose magnitude exceeds those of smaller changes that are still meaningful to market participants. Minimum order size may be too high (not common).
+- **Just right:** Every amount of asset is expressible as a size with an equivalent asset amount that's close enough to the original amount so as to not change its value by a meaningful amount.
+
+Let's create a market with these key properties in mind, and find a configuration that makes sense in light of these concerns.
+
+### Creating a hypothetical market
+
 Consider a hypothetical trading pair `APT/USDC`, `APT` denominated in `USDC`.
 Here, `APT` is considered the "base" asset and `USDC` is considered the "quote asset", meaning that orders sized in `APT` are quoted in `USDC` per `APT`:
 
-| Term        | Specifies   | Symbol |
-| ----------- | ----------- | ------ |
-| Base asset  | Order size  | `APT`  |
-| Quote asset | Order price | `USDC` |
+| Term        | Specifies   | Symbol | Unit Decimals | Unit Price |
+| ----------- | ----------- | ------ | ------------- | ---------- |
+| Base asset  | Order size  | `APT`  | 8             | \$7.32     |
+| Quote asset | Order price | `USDC` | 6             | \$1.00     |
 
-In addition to a base/quote trading pair, each market in Econia additionally contains the following parameters, which are selected by the market registrant during registration:
+We'd like to find a market configuration that makes sense given the above economic state.
+Configuring a market requires three values beyond the base and quote type:
 
-| Parameter          | Meaning                     | Example     |
-| ------------------ | --------------------------- | ----------- |
-| Lot size           | Order size granularity      | 0.1 `APT`   |
-| Tick size          | Price granularity           | 0.01 `USDC` |
-| Minimum order size | Smallest allowed order size | 0.5 `APT`   |
-
-On this market, an order for 7.8 `APT` at a price of 5.23 `USDC` per `APT` would be valid, but the following would be invalid:
-
-| Size | Price | Reason             |
-| ---- | ----- | ------------------ |
-| 7.85 | 5.23  | Size too granular  |
-| 7.8  | 5.235 | Price too granular |
-| 0.4  | 5.23  | Size too small     |
-
-Econia's matching engine uses integers rather than decimals, such that conversion from decimal amounts to integer amounts requires the `aptos_framework::coin::CoinInfo.decimals` for a given `CoinType`:
-
-| Term           | Coin   | Amount |
-| -------------- | ------ | ------ |
-| Base decimals  | `APT`  | 8      |
-| Quote decimals | `USDC` | 6      |
-
-Here, a decimal amount of coins (e.g. 7.8 `APT`), can be converted to an integer amount of indivisible coin subunits (`aptos_framework::coin::Coin.value`):
-
-```python
->>> from decimal import Decimal as dec
->>> decimals = 8
->>> coins = dec('123.456')
->>> subunits = int(coins * 10 ** decimals)
->>> subunits
-12345600000
-```
-
-Similarly, other terms can be converted as follows:
-
-| Variable                   | Units           | Description                                      |
-| -------------------------- | --------------- | ------------------------------------------------ |
-| Lot size                   | Subunits        | Subunits of base (granularity of base amounts)   |
-| Tick size                  | Subunits        | Subunits of quote (granularity of quote amounts) |
-| Minimum order size         | Lots            | Minimum number of lots in an order               |
-| Size                       | Lots            | Number of lots an order is worth                 |
-| Price                      | Ticks (per lot) | Number of ticks one lot is worth                 |
-| Total quote amount to fill | Subunits        | Value of whole order in quote subunits           |
+| Parameter          | Meaning                  | Units                      |
+| ------------------ | ------------------------ | -------------------------- |
+| Lot size           | Order size granularity   | Subunits of base (`APT`)   |
+| Tick size          | Lot value granularity    | Subunits of quote (`USDC`) |
+| Minimum order size | Minimum limit order lots | Number of lots             |
 
 This section utilizes helpers from the Econia Python SDK.
-If you'd like to follow along, run:
+If you'd like to follow along, run in the nearest terminal:
 
 ```bash
 pip3 install econia-sdk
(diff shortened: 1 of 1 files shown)


doc/doc-site/docs/overview/orders.md
```diff
@@ -592,14 +563,19 @@ Hence decimal tick size must be reduced so that integer prices can fit into a 32
...
 Still, however, there are several guidelines that can aid the process:
 
+1. **One tick should be small in value relative to one lot**:
+   since the minimum expressible price is 1 tick per lot, followed by 2, 3 and so on, it's ideal for these prices to be small relative to the value of one lot.
+
 1. **Only specify as much granularity as needed**:
    overly-granular lot size and tick size combinations do not properly translate to integer tick sizes, and may make indexing more difficult.
    Three orders at price 12.34 is easier to interpret than one order each at 12.3401, 12.3404, and 12.3407.
+   In addition, note that lowering the tick size also decreases the maximum expressible price (always worth $2^{32}-1$ ticks).
+   There is a Python helper, among those above, for calculating the maximum price in decimal unit terms: `get_max_quote_per_base_nominal` (see `get_min_quote_per_base_nominal`).
```
alnoki, line 573: I suggest including examples of these helpers above:

* `get_min_quote_per_base_nominal` for the first example under "Units and market parameters"
* `get_max_quote_per_base_nominal` under the wBTC examples

Then in this section probably reference how the SDK functions are used in the walkthrough and talk about playing around with them more

Ideally this will link to the Python SDK page with a relative doc link (something like ../off-chain/python-sdk)

doc/doc-site/docs/overview/orders.md
```diff
@@ -259,110 +259,81 @@ Here, a decimal amount of coins (e.g. 7.8 `APT`), can be converted to an integer
...
    ```
-1. $$\LARGE s_d = s_i l_i 10 ^ {-d_b}$$
-   ```python
-   >>> size_decimal = dec(size_integer) * lot_size_integer / 10 ** decimals_base
-   >>> size_decimal
-   Decimal('7.8')
-   ```
-1. $$\LARGE p_i = \frac{p_d}{t_d}$$
+
+1. Price (ticks per lot)
```
alnoki, line 320: Here it might be useful to show min and max quote per base

alnoki (changes requested):
Just a few minor reviews

Also I noticed that when I originally wrote this I used 2^32 instead of 2^32-1 in some places, how about fixing that here too?

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
 
-In addition to a base/quote trading pair, each market in Econia additionally contains the following parameters, which are selected by the market registrant during registration:
+Market participants should be able to express prices that are close enough together to be effectively continuous while being far enough apart to reflect meaningful movement with a single increment.
 
-| Parameter          | Meaning                     | Example     |
-| ------------------ | --------------------------- | ----------- |
-| Lot size           | Order size granularity      | 0.1 `APT`   |
-| Tick size          | Price granularity           | 0.01 `USDC` |
-| Minimum order size | Smallest allowed order size | 0.5 `APT`   |
+- **Too granular:** Almost every order occupies its own price level, and user interfaces struggle to display a useful segment of the order book before running out of visual real-estate. Maximum price may be too low (not common).
```
alnoki, line 224: Should line break after periods, which is possible in lists:

```
- The first sentence.
  The second sentence.
- A new bullet point.
```

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
-| ---- | ----- | ------------------ |
-| 7.85 | 5.23  | Size too granular  |
-| 7.8  | 5.235 | Price too granular |
-| 0.4  | 5.23  | Size too small     |
+Market participants should be able to create a limit order as long as it's for a meaningful amount of asset.
+Each individual limit order affected by a market order adds to the network fee cost of the market order's execution.
+Thus, it's in every participant's interest that every limit order has a meaningful enough size to include in the execution of a market order.
 
-Econia's matching engine uses integers rather than decimals, such that conversion from decimal amounts to integer amounts requires the `aptos_framework::coin::CoinInfo.decimals` for a given `CoinType`:
+- **Too low:** The limit order books become polluted with small orders and it becomes difficult or impossible to execute a market order without paying a significant network fee. Denial-of-Service is a concern at this point because it becomes possible to "shut down" trading by spamming limit orders.
```
alnoki, line 234: nit: DoS is listed below in a different section, and is not capitalized. Hence here should be "Denial-of-service", but also I suggest changing anyways to

"Malicious actors may conduct a denial-of-service attack by sending in lots of orders with negligible size."

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+- **Too high:** Some or many meaningful market participants are cut out of trading action and thus must take their liquidity somewhere else.
+- **Just right:** Nobody with good intentions is cut out of liquidity provision based on the amount of asset available to them, as long as their amount is meaningful.
```
alnoki, line 236: "Anyone who wants to provide liquidity can, even at smaller amounts, but malicious actors are denied an attack vector"

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
 
-```python
->>> from decimal import Decimal as dec
->>> decimals = 8
->>> coins = dec('123.456')
->>> subunits = int(coins * 10 ** decimals)
->>> subunits
-12345600000
-```
+- **Too granular:** Price granularity is too coarse due to lack of ability to express the price of one size increment (or decrement) in terms smaller than a single subunit of asset.
```
alnoki, line 243: Typo: "Price granularity" should be size granularity

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
-1
+>>> tick_size_unit = "0.00001"
+>>> quote_decimals = 6
+>>> get_tick_size_integer(tick_size_unit, quote_decimals)
+10
 ```
 
-A decimal tick size of 0.001 `USDC` per `APT` would not be supported, however, because at the lowest possible price of 1 tick, the increment in total quote amount for each additional lot could not be represented as an integer multiple of subunits:
+It's easy to tell if this lot size and tick size combination is reasonable by checking the minimum and maximum price under these conditions in nominal (unit) terms.
+Note that the minimum price is also equivalent to price granularity--this tells us the granularity of value for a whole unit of `APT` as opposed to just one lot (equal here to 100000 subunits of `APT`).
```
alnoki, line 301: `--` would ideally render as an m-dash but I'm guessing not here, and also maybe just change for more specificity:

"Note that the minimum price is also equivalent to price granularity, which tells us the granularity of value for a whole unit of APT as opposed to just one lot..."

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
->>> (quote_increment_decimal * 10 ** decimals_quote).normalize()
-Decimal('0.1')
+>>> get_min_quote_per_base_nominal(lot_size_unit, tick_size_unit)
+0.01
+>>> get_max_quote_per_base_nominal(lot_size_unit, tick_size_unit)
+42949672.95
 ```
 
-Hence to check that a lot size/tick size combination is even possible for a market:
+It looks like our price granularity is one penny and our maximum price is "plenty" high.
```
alnoki, line 311: How abou we take "plenty" out of quotes, since they aren't needed here? Indeed it is plenty high

Also, could you illustrate what this means: if we configure a market this way, it will support APT prices up to $42M USDC per APT

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+>>> get_max_quote_per_base_nominal(lot_size_unit, tick_size_unit_new)
+4294967.295
+```
 
-   ```python
-   >>> to_check = lot_size_decimal * 10 ** decimals_base
-   >>> lot_size_integer = int(to_check)
-   >>> assert to_check >= 1 and to_check == lot_size_integer
-   ```
+Notice the maximum price has gone down by one order of magnitude, but it's still "plenty" high.
```
alnoki, line 333: I suggest no quotes around plenty, and maybe be more concrete (e.g. supports APT prices up to $4.2 M per APT)

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+We're almost done at this point: we just need a minimum order size.
+Let's require a minimum limit order value of \$50, that is ~6.83 `APT`, equal to $\$50 / \$7.32$.
```
alnoki, line 335: While this is arithmetically valid, I suggest we use something like 0.5 APT, since the example here is probalby going to be viewed as the canonical market

E.g. we want people to be able to do swaps for less than $50, and also if APT price doubles or triples 0.5 APT will still be valid

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
 
 ### Noteworthy examples
 
 Consider the trading pair `wBTC/USDC`:
-as of the time of this writing, one `BTC` costs approximately 17792.27 `USD`, so initial choices for lot size and tick size might entail:
+as of the time of this writing, one `BTC` costs approximately 17792.27 `USD`, so initial choices for lot size and tick size might be:
 
-1. 0.00001 `wBTC` decimal lot size (corresponding to 0.1779227 `USDC` nominal).
+1. 0.00001 `wBTC` decimal lot size (a value of 0.1779227 `USDC`).
 1. 0.01 `USDC` decimal tick size.
```
alnoki, line 362: I think this example as originally written (by me) was confusing, and I suggest taking out the term "decimal tick size"

Also, `wBTC` should be 9 decimals, and given that it isn't mentioned up front I suspect this example needs to be rewritten

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+
+If we increase lot size by a factor of 10 to 0.0001 `WBTC` which corresponds to an order value granularity of \$1.779 instead of \$0.1779, and keep the above tick size, then we have a price granularity of one penny:
 
-Note, however, that compared with the other viable parameter set, this set trades off size precision for price precision:
-twice as much size granularity entails one half as much price granularity.
+```python
+>>> lot_size = "0.0001" # changed by * 10
+>>> tick_size = "0.000001"
+>>> get_min_quote_per_base_nominal(lot_size, tick_size)
+0.01
```
alnoki, line 424: I think this example will be found confusing by the follower unless it culminates in actual lot size, tick size, min size amounts

Can you show how these would be calculated, using the 9 decimals for wBTC and 6 for USDC?

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
->>> assert to_check >= 1 and to_check == tick_size_integer
->>> tick_size_integer
-1
 ```
+>>> get_tick_size_integer("0.0000001", 6)
+Traceback (most recent call last):
+  File "<stdin>", line 1, in <module>
+  File "/Users/egd/projects/econia/src/python/sdk/econia_sdk/utils/decimals.py", line 46, in get_tick_size_integer
+    return get_lot_size_integer(smallest_decimal_unit, coin_decimals)
+  File "/Users/egd/projects/econia/src/python/sdk/econia_sdk/utils/decimals.py", line 31, in get_lot_size_integer
```
alnoki, line 411: I don't think it's good practice to include personal user names here

And most people stop reading the moment they see a traceback, so I think it's easiest to just show:

```python
>>> get_tick_size_integer("0.0000001", 6)
Traceback (most recent call last):
...
ValueError: Decimal unit too small to represent with 1 subunit
```

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+If we increase lot size by a factor of 10 to 0.0001 `WBTC` which corresponds to an order value granularity of \$1.779 instead of \$0.1779, and keep the above tick size, then we have a price granularity of one penny:
 
-Note, however, that compared with the other viable parameter set, this set trades off size precision for price precision:
-twice as much size granularity entails one half as much price granularity.
+```python
+>>> lot_size = "0.0001" # changed by * 10
+>>> tick_size = "0.000001"
+>>> get_min_quote_per_base_nominal(lot_size, tick_size)
+0.01
+```
```
alnoki, line 426: The below note is out of date due to due to changes in market order APIs that never got documented here, and it is probably not needed. Will you please take it out?

doc/doc-site/docs/overview/orders.md
```diff
@@ -514,98 +431,91 @@ Market orders and swaps allow users to specify precise trade amounts down to a s
 
 :::
 
-Hence the nominal price of 17792.27 `USD` per `BTC` must be truncated to either 17792.26 or 17792.28.
-For the latter case, this entails an integer price of 889614:
+Next, consider a liquid staking derivative `sAPT` trading against `APT`, `sAPT/APT`, both having 8 decimals.
+Here, very-high price granularity is necessary because the pair is "mean reverting" (to a ratio of 1):
```
alnoki, line 435: How about?

"Since the two assets have almost the same nominal price, price granularity will need to be higher here"

doc/doc-site/docs/overview/orders.md
```diff
@@ -514,98 +431,91 @@ Market orders and swaps allow users to specify precise trade amounts down to a s
...
-| ----------------- | ------------------------- |
-| Decimal lot size  | 0.01 `sAPT`               |
-| Decimal tick size | 0.000001 `APT` per `sAPT` |
-| Base decimals     | 8                         |
-| Quote decimals    | 8                         |
-| Decimal price     | 1.000012 `APT` per `sAPT` |
+This gives us a price precision of 1/100th of a penny, with room to go even more precise should we desire.
+However we recall that the most granular possible price isn't necessarily desirable due to almost every order then occupying its own price level, cluttering user interfaces.
+Some coarseness is desirable for that reason.
+Although it is possible for user interfaces to summarize orders into price levels that don't actually exist, this would diverge what the user sees from reality in a way that can't be guaranteed safe.
```
alnoki, line 458: "this approach would represent to the user information that is not accurate, and is not recommended"

doc/doc-site/docs/overview/orders.md
```diff
@@ -514,98 +431,91 @@ Market orders and swaps allow users to specify precise trade amounts down to a s
...
+Should we use the maximum precision of 0.00000001 `APT` then we observe that the maximum price lowers but remains high enough to facilitate this "mean reverting" pair:
```
alnoki, line 459: I suggest taking out mean reverting as most people won't know what this means

"but remains high enough for this pair, where the assets have almost the same price"

doc/doc-site/docs/overview/orders.md
```diff
@@ -215,297 +215,214 @@ Econia does not use a critical height of 1. The actual critical height is define
...
+>>> tick_size = "0.00001" # changed by / 1000
+>>> get_min_quote_per_base_nominal(lot_size, tick_size)
+1.0
+>>> get_max_quote_per_base_nominal(lot_size, tick_size)
+4294967295.0
 ```
 
-However, the new decimal order size granularity corresponds to 1.779227 `USDC` nominal, which may not be considered granular enough.
-Hence alternative parametric inputs might entail:
+That's a minimum price (and price granularity) of \$1, as well as a maximum price that's "plenty" high.
```
alnoki, line 390: "plenty" -> plenty

alnoki (changes requested):
Looks much better now with the rewrites

See assorted comments

alnoki (approved):

PR #422, opened 2023-08-09
[ECO-228] Update Rust SDK

CRBl69:
Core changes:

- Add missing entry functions:
    - `init_market_event_handles_if_missing`.
    - `place_market_order_user_entry`.
- Add `view.rs` file which contains:
    - a struct named `EconiaViewClient` which can query every view function of the contract. Some functions which results can be calculated without blockchain data are calculated locally. Functions which are calculated locally instead of reaching out to the blockchain are identifiable through a note in the function's documentation comment.
    - structs which represent some view functions return type.
- Add missing event types:
    - `MakerEvent`.
    - `TakerEvent`.
- Update old events which did not deserialize correctly (`base_*` and `quote_*` replaced with `base_type` and `quote_type`).
- Re-export `econia_types` as `types`.
- Add `get_events_by_creation_number` helper function.
- Add/update order related types and constants.

### DIFF
--- a/src/rust/sdk/src/entry.rs
+++ b/src/rust/sdk/src/entry.rs
@@ -216,7 +216,7 @@ pub fn change_order_size_user(
 /// * `base`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for base coin.
 /// * `quote`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for quote coin.
 /// * `market_id`: Market ID for corresponding market.
-/// * `integrator`: Integrator's AccountAddress.
+/// * `integrator`: Integrator's `AccountAddress`.
 /// * `side`: Order [`Side`].
 /// * `size`: Size of the order in lots.
 /// * `advance_style`: The [`AdvanceStyle`] of the order.
@@ -600,7 +600,7 @@ pub fn init_market_event_handles_if_missing(
     );
     Ok(EntryFunction::new(
         module,
-        ident_str!("init_market_event_handle_if_missing").to_owned(),
+        ident_str!("init_market_event_handles_if_missing").to_owned(),
         vec![],
         vec![bcs::to_bytes(&market_id)?, bcs::to_bytes(&custodian_id)?],
     ))
--- a/src/rust/sdk/src/entry.rs
+++ b/src/rust/sdk/src/entry.rs
@@ -582,3 +582,66 @@ pub fn withdraw_to_coinstore(
         vec![bcs::to_bytes(&market_id)?, bcs::to_bytes(&amount)?],
     ))
 }
+
+/// Create the `EntryFunction` for [init_market_event_handle_if_missing](https://github.com/econia-labs/econia/blob/main/src/move/econia/doc/user.md#0xc0deb00c_user_init_market_event_handles_if_missing)
+///
+/// Arguments:
+/// * `econia_address`: Aptos `AccountAddress` of the account that holds the econia modules.
+/// * `market_id`: Market ID for corresponding market.
+/// * `custodian_id`: ID of market custodian.
+pub fn init_market_event_handles_if_missing(
+    econia_address: AccountAddress,
+    market_id: u64,
+    custodian_id: u64,
+) -> EconiaResult<EntryFunction> {
+    let module = ModuleId::from(
+        MoveModuleId::from_str(&format!("{}::user", econia_address))
+            .map_err(|a| EconiaError::InvalidModuleId(a.to_string()))?,
+    );
+    Ok(EntryFunction::new(
+        module,
+        ident_str!("init_market_event_handle_if_missing").to_owned(),
+        vec![],
+        vec![bcs::to_bytes(&market_id)?, bcs::to_bytes(&custodian_id)?],
+    ))
+}
+
+/// Create the `EntryFunction` for [place_market_order_user_entry](https://github.com/econia-labs/econia/blob/main/src/move/econia/doc/user.md#0xc0deb00c_market_place_market_order_user_entry)
+///
+/// Arguments:
+/// * `econia_address`: Aptos `AccountAddress` of the account that holds the econia modules.
+/// * `base`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for base coin.
+/// * `quote`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for quote coin.
+/// * `market_id`: Market ID for corresponding market.
+/// * `integrator`: Integrator's AccountAddress.
+/// * `side`: Order [`Side`].
+/// * `size`: Size of the order in lots.
+/// * `restriction`: the [`Restriction`] of the order.
+/// * `self_match_behavior`: the [`SelfMatchBehavior`] of the order.
+pub fn place_market_order_user_entry(
+    econia_address: AccountAddress,
+    base: &TypeTag,
+    quote: &TypeTag,
+    market_id: u64,
+    integrator: &AccountAddress,
+    side: Side,
+    size: u64,
+    self_match_behavior: SelfMatchBehavior,
+) -> EconiaResult<EntryFunction> {
+    let module = ModuleId::from(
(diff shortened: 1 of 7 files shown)


alnoki (changes requested):
@CRBl69 It looks like there are merge conflicts

elliottdehn:
By convention, we don't do empty diff descriptions because they're public (and helpful)! Could you please add a description with an inventory of what you've added (and perhaps why) so that it's easier to review? Will go through now without, but a review is more productive with a good description. Thanks.

src/rust/sdk/src/entry.rs
```diff
@@ -582,3 +582,66 @@ pub fn withdraw_to_coinstore(
...
+}
+
+/// Create the `EntryFunction` for [place_market_order_user_entry](https://github.com/econia-labs/econia/blob/main/src/move/econia/doc/user.md#0xc0deb00c_market_place_market_order_user_entry)
+///
+/// Arguments:
+/// * `econia_address`: Aptos `AccountAddress` of the account that holds the econia modules.
+/// * `base`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for base coin.
+/// * `quote`: Aptos [`TypeTag`](https://docs.rs/move-core-types/0.0.3/move_core_types/language_storage/enum.TypeTag.html) for quote coin.
+/// * `market_id`: Market ID for corresponding market.
+/// * `integrator`: Integrator's AccountAddress.
```
alnoki, line 616: Missing backticks: `AccountAddress`

src/rust/sdk/src/view.rs
```diff
@@ -0,0 +1,1128 @@
...
+            .ok_or(EconiaError::InvalidResponse)?;
+        Ok(value)
+    }
+
+    /// Call `get_price_levels` view function.
+    ///
+    /// Arguments:
+    /// * `market_id`: the id of the market to get the price levels from.
+    /// * `n_ask_levels_max`: the maximum number of ask price levels to index.
+    /// * `n_bid_levels_max`: the maximum number of bidsprice levels to index.
```
alnoki, line 373: change "bidsprice" to "bid price"

src/rust/sdk/src/view.rs
```diff
@@ -0,0 +1,1128 @@
...
+                .clone(),
+        )?;
+        Ok(value)
+    }
+
+    /// Call `get_posted_order_id_side` view function.
+    ///
+    /// Arguments:
+    /// * `user`: the address of the user to check market account ids for.
+    /// * `market_id`: the id of the market to check market accounts for.
```
alnoki, line 342: I'll note that this function also reduces to bitwise operations that can be computed off-chain (only because you were doing bitwise rather than chain calls for other such view funcs), not that you need to update it to the bitwise operations

src/rust/sdk/src/view.rs
```diff
@@ -0,0 +1,1128 @@
...
+            .as_bool()
+            .ok_or(EconiaError::InvalidResponse)?;
+        Ok(value)
+    }
+
+    // Registry related view functions <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+    // User related view functions >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+    /// Call `get_all_market_account_ids_for_market_id` view fuction.
```
alnoki, line 763: typo: "function"

src/rust/sdk/src/view.rs
```diff
@@ -0,0 +1,1128 @@
...
+    T::Err: Display,
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    T::from_str(&s).map_err(serde::de::Error::custom)
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MarketEventHandleCreationInfo {
+    pub ressource_account_address: AccountAddress,
```
alnoki, line 33: This should be `resource_account_address` (currently it says "ressource")

src/rust/types/src/events.rs
```diff
@@ -41,22 +42,80 @@ pub enum CancelReason {
...
+    serde(rename_all = "snake_case")
+)]
+pub enum MakerEventType {
+    Cancel,
+    Change,
+    Evict,
+    Place,
+}
+
+impl TryFrom<u8> for MakerEventType {
```
alnoki, line 58: Note that `MakerEvent` and `TakerEvent` are deprecated (see market.move or the changelog: https://econia.dev/move/changelog)

src/rust/types/src/events.rs
```diff
@@ -41,22 +42,80 @@ pub enum CancelReason {
...
+    pub market_order_id: u128,
+    pub maker: String,
+    pub custodian_id: Option<u64>,
+    pub size: u64,
+    pub price: u64,
+}
+
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Type {
```
alnoki, line 102: How about switching this to `TypeInfo` to be more consistent with the Move type?

src/rust/types/src/order.rs
```diff
@@ -156,6 +190,72 @@ pub enum OrderState {
...
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            1 => Ok(CancelReason::Eviction),
+            2 => Ok(CancelReason::ImmediateOrCancel),
+            3 => Ok(CancelReason::ManualCancel),
+            4 => Ok(CancelReason::MaxQuoteTraded),
+            5 => Ok(CancelReason::NotEnoughLiquidity),
+            6 => Ok(CancelReason::SelfMatchMaker),
+            7 => Ok(CancelReason::SelfMatchTaker),
+            8 => Ok(CancelReason::TooSmallAfterMatching),
```
alnoki, line 251: The last few reasons are out of date, it should be:

* 8: TooSmallToFillLot
* 9: ViolatedLimitPrice

alnoki (changes requested):
Looks mostly good so far, there are just a few typo/struct consistency modifications as well as some enum corrections

elliottdehn:
> Added missing event types: MakerEvent, TakerEvent

I believe these are not necessary to add as inspecting the Move code with CTRL + F indicates these events aren't created anywhere by anything at least in market.move. Also this:

```
        /// Deprecated field retained for compatible upgrade policy.
        maker_events: EventHandle<MakerEvent>,
        /// Deprecated field retained for compatible upgrade policy.
        taker_events: EventHandle<TakerEvent>
````

**Review by elliottdehn — CHANGES_REQUESTED** [review]

File: src/rust/sdk/src/errors.rs [hunk 9]
```diff
@@ -44,6 +112,12 @@ pub enum EconiaError {
     #[error("failed submitting aptos transaction")]
     FailedSubmittingTransaction,
 
+    #[error("invalid response from the contract")]
+    InvalidResponse,
```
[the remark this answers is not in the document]
CRBl69: IDK, if using an older SDK with a newer version of the contract (or the other way around), you could have an invalid response (if the response structure changes). I thought this could be the enum variant for that.

src/rust/sdk/src/view.rs
CRBl69: I missed that. I saw the call to some functions in it and thought it was something more complex. I did it in Rust as well now.

src/rust/types/src/order.rs
CRBl69: Seems that this is duplicated, and this type is already in another place. Went away and removed it all.

src/rust/sdk/src/errors.rs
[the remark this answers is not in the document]
elliottdehn: It's no big deal, however, it's impossible for future versions of the contract to break this code due to the Compatible Upgrade Policy which means fields can't be removed from structs for example, and entry function signatures can't change.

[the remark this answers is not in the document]
alnoki: I think that `InvalidResponse` is fine, as it looks like this error is getting used for view functions

The Aptos compatibility guarantees are weak in the sense that they only check ABI modifications, so it is possible to send back a new data schema that technically doesn't break the ABI but is semantically invalid nonetheless

Like if a field is a vec of `u8` that is supposed to represent a hash digest, but the field is deprecated and the field just comes back empty

alnoki (approved):

elliottdehn (approved):

PR #428, opened 2023-08-17
[ECO-378] Remove unnecessary mutable borrows in view funcs

alnoki:
@chen-robert

This PR removes unnecessary mutable borrows in view functions.

Note that https://github.com/aptos-labs/aptos-core/pull/9369 changed an event unit testing function signature, and so tests are updated here to enable a test run.

All tests pass.

PR #429, opened 2023-08-17
[ECO-379] Check order price field against lookup key

alnoki:
@chen-robert

This PR adds a check for an order's price field against its corresponding AVL queue insertion key, during price level computation.

A failure test is included.

This PR is branched from #428, and should be merged after it merges

### DIFF
--- a/src/move/econia/sources/market.move
+++ b/src/move/econia/sources/market.move
@@ -851,6 +851,8 @@ module econia::market {
     const E_SIZE_CHANGE_INSERTION_ERROR: u64 = 30;
     /// Order ID corresponds to an order that did not post.
     const E_ORDER_DID_NOT_POST: u64 = 31;
+    /// Order price field does not match AVL queue insertion key price.
+    const E_ORDER_PRICE_MISMATCH: u64 = 32;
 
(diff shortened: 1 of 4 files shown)


src/move/econia/sources/market.move
```diff
@@ -2839,10 +2842,16 @@ module econia::market {
...
+                        // Pop order, storing only its size and price.
+                        let Order{
+                            size: order_size,
+                            price: order_price,
+                            user: _,
+                            custodian_id: _,
+                            order_access_key: _
+                        } = avl_queue::pop_head(avlq_ref_mut);
+                        // Verify order price equals insertion key.
+                        assert!(order_price == price, E_ORDER_PRICE_MISMATCH);
```
elliottdehn, line 2854: Why add this here? Seems like something that should/would go everywhere we call `pop_head` (or even anywhere we get an order at all). Basically what makes this location special.

alnoki: First I'll note that the codebase doesn't have any runtime logic for mutating an order's price field once it has been submitted

The other code already was already evaluated as not a security risk, and this only offered "as a small suggestion" from an auditor on new code

elliottdehn: Okay, i see.

elliottdehn (approved):

PR #431, opened 2023-08-18
[ECO-343] Add local testnet dockerfile for e2e testing

alnoki:
This PR adds a local testnet dockerfile for e2e testing, with a README.

I tried installing via CLI script before settling on the `brew` install methodt: https://aptos.dev/tools/aptos-cli/install-cli/automated-install/

The script install method ended up throwing errors about the environment platform, and forced me to use a `--platform` specification during the build phase. I had difficulties trying to modify the path in the image after the scripted install, so I settled on the brew methodology here, which essentially works out of the box albeit with longer build times and an image that is larger than it has to be.

An alternative version of this Dockerimage would involve a `--release` cargo build of the Aptos CLI from source from within the container, such that only the platform-specific binary file ends up in the image. However this would involve more steps and would likely still involve longer build times. 

Given that this local testnet is only for e2e testing, I am not concerned with the potential performance disadvantages that may come with a brew install inside of an emulated environment (Docker container)

### DIFF
--- a/src/rust/e2e/local-testnet/Dockerfile
+++ b/src/rust/e2e/local-testnet/Dockerfile
@@ -1,0 +1,5 @@
+FROM homebrew/brew:latest
+
+RUN brew install aptos
+
+CMD aptos node run-local-testnet --with-faucet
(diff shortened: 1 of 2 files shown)


elliottdehn:
From offline discussion:

This is not that useful because it doesn't have the faucet and exchange contracts deployed. Part of what these e2e testing frameworks need is a node that knows what Econia and the Faucet are, which requires to deploy the source code, which is the hard part of this and why we want docker.

alnoki:
Closing in lieu of #434

PR #434, opened 2023-08-19
[ECO-375] Create dockerized local testnet with Econia and Econia Faucet

alnoki:
This PR adds:

1. A Dockerfile for building the `aptos` CLI from source
2. A Dockerfile for using (1) as a base image to deploy Econia and the Econia faucet under a local testnet
3. Compromised private keys to publish Econia and the Econia faucet under vanity addresses
4. A README with commands showing how to build, run, and do a basic test of the eventual local testnet container


src/docker/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```bash
+CONTAINER_ID=$(docker run \
+    --detach \
+    --publish 8080:8080 \
+    --publish 8081:8081 \
+    local-testnet)
+```
+
+docker run local-testnet
```
elliottdehn, line 60: I think this was supposed to be in the code block

src/docker/local-testnet/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+        --assume-yes \
+        --included-artifacts none \
+        --named-addresses econia_faucet=$(cat accounts/faucet.address) \
+        --override-size-check \
+        --package-dir faucet \
+        --profile faucet
+
+# Copy over chain data only.
+FROM aptos-cli as runtime
+COPY --from=publish /app/data /app/data
```
elliottdehn, line 54: why not have

```
EXPOSE 8080
EXPOSE 8081
```

alnoki: As mentioned on the Dockerfile reference for `EXPOSE`: https://docs.docker.com/engine/reference/builder/#expose

```
The EXPOSE instruction does not actually publish the port.
```

src/docker/README.md
alnoki: Nice catch, the code block actually handles this.

d01db6a removes the dangling command

src/docker/local-testnet/Dockerfile
alnoki: Noting here discussion at https://stackoverflow.com/questions/22111060

08ddde3 adds requested `EXPOSE` commands

elliottdehn: TIL about publish vs expose

elliottdehn (approved):

PR #439, opened 2023-08-22
[ECO-409] Create docker compose with indexer and postgres

alnoki:
This PR adds a docker compose environment with an indexer, local testnet, and postgres instance.

It contains an updated README with instructions on running the environment, as well as assorted docker commands that are useful during development.

The genesis and waypoint files for the local chain are stored in the indexer image under `/app/`. Note that if you want to inspect the files run the compose command from the README then use the command at the bottom to inspect the data directory using a shell inside the image.

Note that an indexer.yaml file will need to be added to the indexer, and the `NODE_YAML` copy line at the end of the indexer dockerfile will need to be uncommented.

The indexer is not yet running/connected to postgres, and will need to have an `ENTRYPOINT` command added at the end.

The indexer submodule has been updated to `econia-labs/aptos-core` under the `main` branch, and the submodule will need to be updated to run indexing operations.


src/docker/README.md
```diff
@@ -1,72 +1,69 @@
...
+  ```
 
-```bash
-docker stop $CONTAINER_ID
-```
+- Prune all containers, images, and volumes:
+
+  ```
+  docker system prune -af --volumes
+  ```
```
CRBl69, line 69: Seems weird to have Docker documentation in our own documentation, but I agree that this can be very helpful for someone that isn't familiar to Docker but wants to use our Docker setup. I would add a link to the official Docker documentation either at the end or at the top of this section.

src/docker/indexer/e2e/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+FROM debian:bookworm-slim AS runtime
+ARG APP_ROOT
+ARG GENESIS_BLOB
+ARG NODE_BIN
+ARG NODE_YAML
+ARG WAYPOINT_TXT
+WORKDIR $APP_ROOT
+COPY --from=builder $NODE_BIN .
+
+# COPY --from=builder $NODE_YAML .
```
CRBl69, line 61: Should this be deleted ?

```suggestion
```

src/docker/README.md
```diff
@@ -1,72 +1,69 @@
...
+# From Econia repo root
+git submodule init
+git submodule update src/rust/dependencies/aptos-core
+```
 
-This Dockerfile builds an image containing the Aptos CLI compiled from source, to avoid potential platform issues that could ensue from other image generation methods.
+## Start up
 
-## Build
+> This command may a while the first time you run it, since it will have to compile the Aptos CLI and an indexer node from source before running several commands against a local testnet.
```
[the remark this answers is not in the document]
alnoki: Updated in f1b55ad

src/docker/README.md
alnoki: Link with explanation added in f1b55ad

src/docker/indexer/e2e/Dockerfile
alnoki: How about the update f1b55ad?

CRBl69 (approved):

PR #442, opened 2023-08-25
[ECO-406] Add new database crate

CRBl69:
Add a new database crate with `market_registration_events` as its only table.

Also added a trigger to push a notification every time a new event is created.


alnoki (changes requested):
Per ECO-406, how about?

1. A docker compose call that initializes the postgres table upon docker compose startup?
2. Some more high-level content in the README?

Also, it looks like there are some merge conflicts against `main`

src/docker/database/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY . .
+
+RUN apt-get update && apt-get install -y \
+    libpq-dev \
+    postgresql-client \
+    && rm -rf /var/lib/apt/lists/*
+RUN cargo install diesel_cli --no-default-features --features postgres
+
+CMD ["diesel", "migration", "run"]
```
alnoki, line 17: How about switching this to `ENTRYPOINT` ?

src/docker/database/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ENV DATABASE_URL=$DATABASE_URL
+WORKDIR $RUST_ROOT/dbv2
+
+COPY . .
+
+RUN apt-get update && apt-get install -y \
+    libpq-dev \
+    postgresql-client \
+    && rm -rf /var/lib/apt/lists/*
+RUN cargo install diesel_cli --no-default-features --features postgres
```
alnoki, line 15: For longer commands like these, can we follow the Dockerfile best practices and line break, alphabetizing arguments?

```
RUN cargo install diesel_cli \
    --features postgres \
    --no-default-features
```

src/docker/database/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# This docker file is used to run the database migrations
+FROM rust:slim-bookworm
+
+ARG RUST_ROOT=src/rust
+ARG DATABASE_URL
+ENV DATABASE_URL=$DATABASE_URL
+WORKDIR $RUST_ROOT/dbv2
+
+COPY . .
```
alnoki, line 9: According to the build context from the docker compose, this will copy the entire econia repostory into the image, and it looks like it will try to do this *after* switching the working directory

I think this block should look something more like

```
# Relative to Econia repository root.
ARG RUST_ROOT=src/rust

ARG DATABASE_URL

ENV DATABASE_URL=$DATABASE_URL
COPY $RUST_ROOT .
WORKDIR dbv2
```

src/docker/compose.e2e.yml
```diff
@@ -36,6 +36,15 @@ services:
     volumes:
       - db:/var/lib/postgresql/data
 
+  diesel:
```
alnoki, line 39: How about alphabetizing the names of the services, such that this is between `chain` and `indexer`?

src/rust/dbv2/README.md
```diff
@@ -0,0 +1,28 @@
+# Database
+
+This repository provides migrations to create the Econia event database, and Rust types to work with the event database.
+
+The aggregator furthermore adds tables to this database under the `aggregator` schema. Those are disucssed in the `src/rust/aggregator` crate.
```
alnoki, line 5: How about a line break after the first sentence?

Also, a typo: `disusssed` -> `discussed`

src/rust/dbv2/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Contributing
+
+When creating a new migration, before committing, please make sure to use [sql-formatter](https://github.com/sql-formatter-org/sql-formatter) with the provided configuration.
+
+Also make sure to add the corresponding documentation to `models.md`.
+
+# Documentation
+
+## diesel
```
alnoki, line 15: If it is going to be lowercase in a header, how about monospace too?

```
## `diesel`
```

src/rust/dbv2/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This crates provides the models for every table in the database.
+
+We also use `diesel` to run migrations. In our docker compose configuration, we have a dedicated docker container that runs `diesel migration run`.
+
+## Notifications
+
+We use [Postgres notifications](https://www.postgresql.org/docs/15/sql-notify.html) to create a real time notification feed for Econia's events.
+
+Each event table has a [trigger](https://www.postgresql.org/docs/15/sql-createtrigger.html) that emits a notification with the name of the event as the channel,
```
alnoki, line 27: The docker compose uses Postgres 14 per the aptos docs, but these links are for postgres 15

Will this be an issue?

src/rust/dbv2/README.md
```diff
@@ -0,0 +1,28 @@
...
+and the raw event data converted to JSON as its payload.
```
alnoki, line 28: How about joining this line with the last one (such that there is no line break after a comma)?

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
+# Models
+
+This document contains a complete list of all the event tables in our database. **If you're searching for aggregator tables, please look into the aggregator crate (`src/rust/aggregator`).**
+
```
alnoki, line 4: How about a line break after the first sentence?

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
...
+## market_registration_event
```
alnoki, line 5: How about monospace in the header?

```
## `market_registration_event`
```

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
...
+
+| Column                  | Description                                                                                                                                  | Type      |
+| ----------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `id`                    | The database ID of the event. Used to chronologically separate events.                                                                       | number    |
+| `market_id`             | The ID of the market just registered.                                                                                                        | number    |
+| `time`                  | Timestamp at which the event happenned.                                                                                                      | timestamp |
```
alnoki, line 11: typo `happenned` -> `happened`

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
+# Models
+
+This document contains a complete list of all the event tables in our database. **If you're searching for aggregator tables, please look into the aggregator crate (`src/rust/aggregator`).**
+
+## market_registration_event
+
+| Column                  | Description                                                                                                                                  | Type      |
+| ----------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
```
alnoki, line 8: How about running this through `mdformat`?

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
...
+| `quote_module_name`     | Module name of the quote token.                                                                                                              | string    |
+| `quote_struct_name`     | Struct name of the quote token.                                                                                                              | string    |
+| `lot_size`              | Number of base units exchanged per lot.                                                                                                      | number    |
+| `tick_size`             | Number of quote coin units exchanged per tick.                                                                                               | number    |
+| `min_size`              | Minimum number of lots per order.                                                                                                            | number    |
+| `underwriter_id`        | [Link](https://github.com/econia-labs/econia/blob/fdab71182d4a7bf1b0fabc1a030e524575442e68/src/move/econia/sources/registry.move#L286-L289). | number    |
```
alnoki, line 22: I'm not sure about adding in these tables, as it looks like a decent amount of duplicate work that isn't autogenerated

e.g. I think that we should rely on a way to update the doc comments in the rust source and have those changes reflected downstream

Is there a way we could use that instead?

Or maybe we just don't even use this file altogether?

The postgres tables should ultimately be abstracted to data consumers from the API, so it seems like this documentation is going to be for developers who will be looking at the source anyways

src/rust/dbv2/models.md
```diff
@@ -0,0 +1,22 @@
...
+| `market_id`             | The ID of the market just registered.                                                                                                        | number    |
+| `time`                  | Timestamp at which the event happenned.                                                                                                      | timestamp |
+| `base_account_address`  | Account address of the base token.                                                                                                           | string    |
+| `base_module_name`      | Module name of the base token.                                                                                                               | string    |
+| `base_struct_name`      | Struct name of the base token.                                                                                                               | string    |
+| `base_name_generic`     | TODO: wut is dis ?                                                                                                                           | string    |
```
alnoki, line 15: https://econia.dev/overview/registry#markets

src/rust/dbv2/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This repository provides migrations to create the Econia event database, and Rust types to work with the event database.
+
+The aggregator furthermore adds tables to this database under the `aggregator` schema. Those are disucssed in the `src/rust/aggregator` crate.
+
+# Contributing
+
+When creating a new migration, before committing, please make sure to use [sql-formatter](https://github.com/sql-formatter-org/sql-formatter) with the provided configuration.
+
+Also make sure to add the corresponding documentation to `models.md`.
```
alnoki, line 11: I left some comments on the `models.md` file, and toward the end of review realized we might want to get right of that file (see comments there)

Essentially it is cleaner if we just point developers to the source or DocGen files, rather than creating duplicate work there to maintain

Hence if we take it out, you miight want to update the description here with a simple reference to the corresponding Rust source

alnoki (changes requested):

src/rust/dbv2/README.md
alnoki: So the `models.md` file has been taken out, but there is still a reference to it here.

How about taking out the reference, here, to `models.md`?

src/docker/database/Dockerfile
alnoki: Per offline discussion, diesel only needs access to the `dbv2` directory tree with the migrations files and the `diesel.toml` config

Additionally it can be somewhat confusing to set `WORKDIR` before the copy operation, because the relevant directory has not yet been copied over

Hence to reduce image caching size and to provide clarity about the working directory, how about?

```Dockerfile
# Relative to Econia repository root.
ARG DATABASE_ROOT=src/rust/dbv2

ARG DATABASE_URL
ENV DATABASE_URL=$DATABASE_URL

COPY $DATABASE_ROOT /db/
WORKDIR /db/
```

src/rust/dbv2/README.md
alnoki: Note that the typo was addressed but not the line break

alnoki (changes requested):
It looks like the last commit didn't fully address all comments above, so I've left them marked unresolved above (I'm guessing that some of them were hidden on your GitHub view?)

Hence per recent PR procedure discussion in eng team call offline, will you please?

1. Address all outstanding comments not marked as resolved, in a single commit (note on the GitHub view some might be hidden so you might need to show hidden ones)
2. Comment on each issue currently marked unresolved, something to the effect of "Addressed in [commit hash]"
3. Leave the issues marked unresolved, for me to follow up on during review

src/docker/database/Dockerfile
CRBl69: Addressed in c7683cb9.

src/rust/dbv2/README.md
CRBl69: Addressed in c7683cb9.

src/rust/dbv2/README.md
CRBl69: Addressed in c7683cb9.

alnoki (approved):

PR #444, opened 2023-08-25
[ECO-420] Add market registration events to aggregator

CRBl69:
Add the aggregator.

`src/rust/aggregator/src/data.rs` contains a trait of what an aggregator datum should be with extensive comment documentation.

`src/rust/aggregator/src/data/markets.rs` is an example implementation of that.

The `main.rs` file creates data aggregators, and periodically runs them.

Migrations were added to create a table to store market registrations per day.

### DIFF
--- a/src/rust/aggregator/src/data/markets.rs
+++ b/src/rust/aggregator/src/data/markets.rs
@@ -1,0 +1,69 @@
+use anyhow::anyhow;
+use chrono::{DateTime, Utc, Duration};
+use sqlx::PgPool;
+
+use super::{Data, DataAggregationResult, DataAggregationError};
+
+pub struct MarketsRegisteredPerDay {
+    pool: PgPool,
+    last_indexed_timestamp: Option<DateTime<Utc>>,
+}
+
+impl MarketsRegisteredPerDay {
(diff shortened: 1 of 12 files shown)


src/rust/aggregator/src/data/markets.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    async fn process_and_save_historical_data(&mut self) -> DataAggregationResult {
+        sqlx::query!(
+            r#"
+                INSERT INTO aggregator.markets_registered_per_day (markets, date)
+                SELECT COUNT(*), time::date
+                FROM market_registration_events
+                WHERE time::date NOT IN (
+                    SELECT date FROM aggregator.markets_registered_per_day
```
elliottdehn, line 35: We should probably not be using timestamp this way. I realize it's unlikely but if there are two events with the same timestamp then we'll miss one. Market registration events have a transaction version number and event index and this comprises a unique identifier. Let's use that.

elliottdehn (changes requested):
Just the timestamp thing

elliottdehn (approved):

PR #449, opened 2023-08-28
[ECO-430] Iindicate which form of address to use in sdk example

CRBl69:


src/rust/sdk/example/src/main.rs
```diff
@@ -61,6 +61,8 @@ aptos move publish \
 
 # run the script
 cargo run -- $APTOS_NODE_URL $APTOS_FAUCET_URL $ECONIA_ADDR $FAUCET_ADDR
+
+# note that when passing addresses to the executable, they must be of the form 0x1234...
```
alnoki, line 65: How about adding a comment to this effect on the Rust SDK markdown page?

E.g. right before "Understanding the example script", something like

:::tip

Make sure to specify your addresses with a leading `0x`, for example `0xabc123...`.

:::

alnoki (approved):

PR #453, opened 2023-08-31
[ECO-450] Add v2 processor submodule

elliottdehn:
We need to fork the v2 processor example repository from Aptos so that we can add to it our new processor, then run it with Docker. Here I've run `git submodule add https://github.com/aptos-labs/aptos-indexer-processors.git` which has added the submodule. I believe that's all I need to do.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "src/typescript/frontend/public/static"]
 	path = src/typescript/frontend/public/static
 	url = git@github.com:tradingview/charting_library.git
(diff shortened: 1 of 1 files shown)


alnoki (changes requested):
Note that this is using Aptos Labs' repo, and if we want to fork it we will have to use as a submodule a fork not yet in our GitHub organization: `econia-labs/aptos-indexer-processors`

As I understand it is TBD if we strictly need to use a submodule or if we can have our own custom processor as a standalone file, so I'll just note that in the former case the changes proposed here won't work because we will need to add our own diff set on top of the fork

PR #457, opened 2023-09-01
ECO-453: query icons regardless if api fails

chiumax:
the reason why it broke is because query icons was grouped in with data api calls but those are broken at the moment.


chiumax:
> Should we always use the icons from `pairData`, but fallback on `coinListClient` when it's unavailable? We can just change the typing to
> 
> ```
>     baseAssetIcon?: string;
>     quoteAssetIcon?: string;
> ```

unsure if this is a related reply, but the asset icons aren't undefined, but they were grouped in with other calls that errored out causing the asset icons be undefined as well

PR #458, opened 2023-09-01
[ECO-454] Integrate aptos-indexer-processors/processor

elliottdehn:
We have added this submodule so that it can be extended with our processor code. However, we have to integrate the various projects within the imported workspace into ours. This will allow it to utilize our types. We have to essentially take over as the workspace for the projects which means adding a bunch of dependencies to our workspace which is probably fine.

I had to be careful because we also have to use the same features, such as `serde_json` now having ordered fields presumably for consistent hashing and such in some code of the processors.

Their workspace Cargo.toml: https://github.com/econia-labs/aptos-indexer-processors/blob/main/rust/Cargo.toml

Processor Cargo.toml: https://github.com/econia-labs/aptos-indexer-processors/blob/main/rust/processor/Cargo.toml
Server-Framework Cargo.toml: https://github.com/econia-labs/aptos-indexer-processors/blob/main/rust/server-framework/Cargo.toml
Protos Cargo.toml: https://github.com/econia-labs/aptos-indexer-processors/blob/main/rust/aptos-indexer-protos/Cargo.toml
Moving-Average Cargo.toml: https://github.com/econia-labs/aptos-indexer-processors/blob/main/rust/moving-average/Cargo.toml

### DIFF
--- a/src/rust/Cargo.toml
+++ b/src/rust/Cargo.toml
@@ -1,30 +1,91 @@
 [workspace]
 resolver = "2"
-members = ["aggregator", "api", "db", "dbv2", "types", "sdk", "sdk/example"]
+members = [
+  "aggregator",
+  "api",
+  "db",
+  "dbv2",
+  "types",
+  "sdk",
+  "sdk/example",
+  "dependencies/aptos-indexer-processors/rust/processor",
+]
 exclude = ["dependencies"]
 
+[workspace.package]
+authors = ["Econia Labs <developers@econialabs.com>"]
+license = "Business Source License"
+homepage = "https://www.econialabs.com/"
+publish = false
+repository = "https://github.com/econia-labs/econia"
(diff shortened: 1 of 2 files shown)


src/rust/Cargo.toml
```diff
@@ -1,30 +1,91 @@
...
+    "serde_json",
+] }
+diesel_migrations = { version = "2.1.0", features = ["postgres"] }
 diesel-derive-enum = "2.1.0"
 dotenvy = "0.15.7"
 envy = "0.4.2"
 field_count = "0.1.1"
+futures = "0.3.24"
 futures-util = "0.3.28"
+gcloud-sdk = { version = "0.20.4", features = [
```
elliottdehn, line 52: these are not perfectly alphabetical because i don't know how to make it sort them

PR #460, opened 2023-09-04
[ECO-414] Add PostgREST

CRBl69:
- Add a `Dockerfile` to start an instance of PostgREST.
- Add a migration to create the `api` schema and an anonymous web user for PostgREST.
- Update DB's README to document the new architecture.


src/rust/dbv2/migrations/2023-09-04-113144_postgrest/up.sql
```diff
@@ -0,0 +1,29 @@
+-- Your SQL goes here
+CREATE ROLE web_anon nologin;
```
elliottdehn, line 2: Maybe drop the `anon` role (and do other undos) that I made in the previous migration.

src/rust/dbv2/migrations/2023-09-04-113144_postgrest/up.sql
```diff
@@ -0,0 +1,29 @@
...
+
+GRANT
+SELECT
+  ON api.market_registration_events TO web_anon;
+
+
+CREATE EXTENSION pgjwt CASCADE;
+
+
+CREATE FUNCTION api.jwt (json) RETURNS TEXT AS $$
```
elliottdehn, line 27: I don't really know what 'econia_is_dope' is doing, and whether or not that's a safe/good way to do the thing.

src/docker/compose.dss-local.yaml
```diff
@@ -44,6 +47,21 @@ services:
...
+      context: ../../
+      dockerfile: src/docker/api/Dockerfile.ws
+      args:
+        - POSTGRES_WEBSOCKETS_VERSION=0.11.1.0
+    depends_on:
+      - diesel
+      - postgres
+    environment:
+      - PGWS_DB_URI=postgres://econia:econia@postgres/econia
+      - PGWS_JWT_SECRET=econia_is_dope
```
elliottdehn, line 61: Doesn't this mean anyone can just give themselves a valid JWT at any time?
I'm trying to understand the purpose behind having JWT at all given that it seems to provide no security for us.

src/docker/compose.dss-local.yaml
```diff
@@ -44,6 +47,21 @@ services:
...
+  ws:
+    build:
+      context: ../../
+      dockerfile: src/docker/api/Dockerfile.ws
+      args:
+        - POSTGRES_WEBSOCKETS_VERSION=0.11.1.0
+    depends_on:
+      - diesel
+      - postgres
+    environment:
```
elliottdehn, line 59: Reading this, it looks like we may want an additional environment variable:
https://github.com/diogob/postgres-websockets#recovering-from-listener-database-connection-failures

> Self healing connection - postgres-websockets comes with a connection supervisor baked in. You just need to set the configuration PGWS_CHECK_LISTENER_INTERVAL to a number of milliseconds that will be the maximum amount of time losing messages. There is a cost for this since at each interval an additional SELECT query will be issued to ensure the listener connection is still active. If the connection is not found, the connection thread will be killed and respawned. This method has the advantage of keeping all channels and websocket connections alive while the database connection is severed (although messages will be lost).

src/docker/compose.dss-local.yaml
```diff
@@ -24,6 +26,7 @@ services:
 
   postgrest:
     depends_on:
+      - diesel
       - postgres
     environment:
       PGRST_DB_URI: "postgres://econia:econia@postgres:5432/econia"
```
elliottdehn, line 32: you'll have to update the role down here to be `web_anon`
The reason this worked is that `anon` is a role that I created that can see anything (see the first migration)

elliottdehn (changes requested):
mainly just the `anon` role stuff to take care of

src/rust/dbv2/migrations/2023-09-04-113144_postgrest/up.sql
CRBl69: The thing is the DB is open. We don't want to issue API keys right now. So I just made a JWT generator that literally gives you full authorizations for querying. A JWT is necessary to use with `postgres-websockets`.

src/docker/compose.dss-local.yaml
CRBl69: Yes exactly, required by the websockets framework.

src/docker/compose.dss-local.yaml
CRBl69: I think they're talking about the `LISTEN` they're using internally to listen for notifications. I added the environment variable with 1000ms as a value. I think it's a reasonable value that has a great balance between not spamming the DB and not having too much downtime.

src/rust/dbv2/migrations/2023-09-04-113144_postgrest/up.sql
```diff
@@ -0,0 +1,29 @@
...
+SELECT
+  *
+FROM
+  market_registration_events;
+
+
+GRANT usage ON SCHEMA api TO web_anon;
+
+
+GRANT
```
elliottdehn, line 19: Add to this:

```
REVOKE SELECT ON ALL TABLES IN SCHEMA public FROM anon;
REVOKE USAGE ON SCHEMA public FROM anon;
DROP ROLE anon;
```

This is the "down" part of the anon role I added, to be removed here.

CRBl69: I removed the part where `anon` is added, so there's no more need for this. I know we discussed not changing old migrations, but this seemed cleaner and since there's no deployment yet, it has no consequences.

elliottdehn (approved):

PR #461, opened 2023-09-04
[ECO-456] Update market parameters in trade.py

elliottdehn:
Right now the parameters are kind of wonky because they were made before we had helpers to go around the tick size intermediary value. Here I've used the `get_market_parameters_integer` helper with a size precision of "0.001" and a price precision of "0.001" along with base decimals 8 and quote decimals 6.

I've run the script through and confirmed it still works with existing values.

### DIFF
--- a/src/python/sdk/examples/trade.py
+++ b/src/python/sdk/examples/trade.py
@@ -482,8 +482,11 @@ def fund_USDC(account: Account, wholes: int):
 
 
 def setup_market(faucet_client: FaucetClient, viewer: EconiaViewer) -> int:
-    lot_size = 10 ** (8 - 3)  # eAPT has 8 decimals, want 1/1000th granularity
-    tick_size = 10 ** (6 - 3)  # eUSDC has 6 decimals, want 1/1000th granularity
+    # Use econia_sdk.utils.decimals.get_market_parameters_integer for these settings.
+    # The settings here give a size precision of 0.001 eAPT and a price precision of
+    # 0.001 eUSDC given eAPT 8 decimals / eUSDC 6 decimals.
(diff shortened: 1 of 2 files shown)


src/python/sdk/examples/trade.py
```diff
@@ -482,8 +482,11 @@ def fund_USDC(account: Account, wholes: int):
 
 
 def setup_market(faucet_client: FaucetClient, viewer: EconiaViewer) -> int:
-    lot_size = 10 ** (8 - 3)  # eAPT has 8 decimals, want 1/1000th granularity
-    tick_size = 10 ** (6 - 3)  # eUSDC has 6 decimals, want 1/1000th granularity
+    # Use econia_sdk.utils.decimals.get_market_parameters_integer for these settings.
+    # The settings here give a size precision of 0.001 eAPT and a price precision of
+    # 0.001 eUSDC given eAPT 8 decimals / eUSDC 6 decimals.
+    lot_size = 100000 
+    tick_size = 1
```
alnoki, line 489: Good call here, can we also update min size to 500 to be consistent with https://econia.dev/overview/orders#creating-hypothetical-markets?

alnoki (changes requested):
Thanks for taking initiative on this

I think updating the min size will be helpful, and since that would already require another commit, could you add in leading `0x` clarification as we did in https://github.com/econia-labs/econia/pull/449?

src/python/sdk/examples/trade.py
elliottdehn: done: a99b7816d360db01025030f898d3ed907db7bb3c

alnoki (approved):

PR #464, opened 2023-09-05
[ECO-413] Add V2 market registration processor

elliottdehn:
This is the humble beginnings of our chain data processor that will put Econia's events in a database. We've started with market registration here and I've done my best to keep things at an absolute minimum.

I had to update our table migration to use a more practical primary key schema as recommended by Aptos, and added a few other things required by the processor. I've only added what the processor would not run without (trial-by-error).

### DIFF
--- a/src/rust/dbv2/migrations/2023-08-25-111710_add_market_registration_events/up.sql
+++ b/src/rust/dbv2/migrations/2023-08-25-111710_add_market_registration_events/up.sql
@@ -1,8 +1,18 @@
 -- Your SQL goes here
 -- Corresponds to econia::registry::MarketRegistrationEvent
+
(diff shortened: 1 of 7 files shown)


src/rust/dbv2/migrations/2023-08-25-111710_add_market_registration_events/up.sql
```diff
@@ -1,8 +1,18 @@
 -- Your SQL goes here
 -- Corresponds to econia::registry::MarketRegistrationEvent
+
+CREATE TABLE ledger_infos (chain_id BIGINT UNIQUE PRIMARY KEY NOT NULL);
```
elliottdehn, line 4: these two tables are yoinked from the migrations that came with the processor project, and were required to run the processor. In the interests of not having two sets of migrations to run I simply included the required ones in our migrations.

PR #465, opened 2023-09-06
[ECO-440] Create processor docker compose

elliottdehn:
I've updated our e2e docker compose to include the processor and postgrest, and added what appears to be a necessary role to the database that can see (but not do) anything. I was able to navigate to https://localhost:3000/ and see our table schemas, as well as https://localhost:3000/market_registration_events to see the one event inserted there by the database.

This means we now have a working API for the events. I'm not sure if we'll use this for public-facing API's since we'll probably be exposing different data objects there. However, knowing that we have postgrest set-up with our database ready for the aggregator to insert into is good.

### DIFF
--- a/src/docker/processor/Dockerfile
+++ b/src/docker/processor/Dockerfile
@@ -1,0 +1,55 @@
+### Indexer Processor Image ###
+
+# Stage 1: Build the binary
+
+FROM rust:slim-bullseye as builder
+
+WORKDIR /app
+
+COPY --link src/rust /app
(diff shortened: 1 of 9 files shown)


src/docker/processor/Dockerfile
```diff
@@ -0,0 +1,55 @@
+### Indexer Processor Image ###
```
elliottdehn, line 1: This was copied (and partly edited for directories) from the dockerfile given to us by the processor project as it is.

alnoki:
Closing in lieu of #466

PR #466, opened 2023-09-06
[ECO-472] Create local DSS compose file

alnoki:
This PR adds integrates an in-review indexer PR that processes market registration events, with instructions for a local data service stack docker compose file that can be run with a simple config file for things like GRPC auth token.


src/docker/processor/config-template.yaml
```diff
@@ -0,0 +1,10 @@
+health_check_port: 8084
+server_config:
+  processor_name: econia_processor
+  postgres_connection_string: postgres://econia:econia@postgres:5432/econia
+  indexer_grpc_data_service_address: <MODIFY_ME>
```
elliottdehn, line 5: This should probably be something unique to each field so that it can be replaced by a build script.

alnoki: @elliottdehn addressed in 86251552

CRBl69 (approved):

elliottdehn (approved):

PR #472, opened 2023-09-13
[ECO-473] Docker compose for TSS

elliottdehn:
This diff adds contract deployments and docker compose entries for TSS.


src/docker/compose.dss-local.yaml
```diff
@@ -2,6 +2,18 @@ version: "3.9"
 
 services:
 
+  streamer:
+    privileged: true
+    build:
+      context: ../../
+      dockerfile: src/docker/tss/Dockerfile
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:18084/readiness"]
```
elliottdehn, line 11: This doesn't work yet: it's 404 despite Aptos' advisement (see chat with them). They're aware and looking into it.
@CRBl69 If you can find another way to health check please do.

src/docker/tss/start.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+dockerd &
+
+sleep 10
+
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
```
elliottdehn, line 7: had to add this for it to work on my machine

src/docker/tss/install_deps.sh
```diff
@@ -0,0 +1,27 @@
...
+  tee /etc/apt/sources.list.d/docker.list > /dev/null
+apt-get update
+
+apt install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin python3 curl gcc python3-dev
+
+rm -rf /var/lib/apt/lists/*
+
+curl -sSL https://install.python-poetry.org | python3 -
+
+curl -L -o g.tar.gz https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz
```
elliottdehn, line 21: @CRBl69 It might be possible to use grpcurl to healthcheck the stream instead of readiness, recommend to try: https://github.com/aptos-labs/aptos-core/blob/main/testsuite/indexer_grpc_local.py#L180-L204

elliottdehn: If we could somehow write a health-check that returns exit code 0 on time-out and anything else otherwise then we'd be golden.

src/docker/tss/start.sh
```diff
@@ -0,0 +1,38 @@
...
+
+poetry run python indexer_grpc_local.py start
+
+aptos init \
```
elliottdehn, line 11: I haven't actually run this bit yet but it's taken straight from /docker/chain (along with other bits) so it should work.

src/docker/compose.dss-local.yaml
CRBl69: image

Seems like it works.

Also, Docker docs say this:

image

When `curl` gets a 404 it seems to exit with code 22:

image

I think you should put `test: ["CMD-SHELL", "curl -f http://localhost:18084/readiness || exit 1"]` here.

src/docker/processor/Dockerfile
```diff
@@ -42,11 +42,21 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     && rm -rf /var/lib/apt/lists/*
 COPY --from=builder $BIN_PATH /usr/local/bin
 
+RUN curl -L -o g.tar.gz https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz
+RUN tar xf g.tar.gz
+RUN mv grpcurl /usr/local/bin
+RUN rm g.tar.gz
+
+WORKDIR /app
+COPY /src/docker/processor/start.sh .
```
alnoki, line 51: Can you install this with `apt-get` as in dockerfile best practices?

src/docker/tss/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+# Relative to image root.
+ARG NODE_BIN=/aptos-core/target/cli/aptos
+
+# Get Aptos repo and build dependencies, then compile.
+FROM rust:1-bookworm AS compile-cli
+ARG GIT_REPO
+ARG NODE_BIN
+RUN git clone \
+    $GIT_REPO \
+    --depth 1
```
alnoki, line 21: This should use a git tag because sometimes the head of `main` is unstable

src/docker/tss/install_deps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+apt-get update
+apt-get install -y --fix-missing git libssh-4 ca-certificates curl
+
+curl -L -o g.tar.gz https://github.com/fullstorydev/grpcurl/releases/download/v1.8.7/grpcurl_1.8.7_linux_x86_64.tar.gz
```
alnoki, line 6: Can we do apt-get for this one?

alnoki (changes requested):
How about updating the README in `src/docker`?

src/docker/tss/Dockerfile
elliottdehn: done

src/docker/processor/Dockerfile
elliottdehn: no, grpcurl is not on apt-get

src/docker/tss/install_deps.sh
elliottdehn: see other comment

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
 
 This Docker compose is designed to work with an end-to-end testing environment, with Econia and the Econia faucet published under single-signer vanity address accounts generated from plaintext (compromised) private keys.
 
-While the local testnet is running, you can look up on-chain Move resources using the published node REST API port (note that the Aptos faucet API may take longer to start up than the node REST API):
+There are a few steps to start up the local end-to-end testing environment:
+
+**1. Configure the processor.**
+
+See the file `config-template-local.yaml` in `/econia/src/docker/processor` for an example configuration.
```
alnoki, line 42: How about taking out `/econia/` so it says `src/docker/processor`?

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+Copy and rename this file to `config.yaml`, which is ignored by git, so that the docker compose will pick it up.
+The new, copied and renamed file goes into the same folder as the template.
+
+**2. Run the docker compose.**
+
+While inside `/econia/src/docker` run:
```
alnoki, line 48: Same as above for taking out `/econia/` which implies an `econia` dir at root

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+```sh
```
alnoki, line 49: How about runinng `mdformat` on this? When I run it it surrounds this with blank lines

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+
+**4. Verify working connectivity** (optional, recommended).
+
+There is a script available that performs most (but not all) of the eventful and/or interesting operations on the exchange.
+By running it against the local end-to-end deployment, it's possible to verify things like processor operation, database connectivity and data accessibility.
+You'll need to install [Poetry](https://python-poetry.org/docs/):
+
+```sh
+brew install poetry
+curl -sSL https://install.python-poetry.org | python3 - # alternative
```
alnoki, line 77: Can we split these into two blocks with the alternative note outside?

Typically people just copy paste and we don't want them having onflicting poetry installs

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+```
+
+Next, navigate your terminal to `/econia/src/python/sdk` and run the following:
```
alnoki, line 80: Same for `/econia/`

src/docker/README.md
```diff
@@ -35,15 +35,79 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+
+```sh
+poetry install
+poetry run trade
+```
+
+The script will have a few prompts; respond as follows:
```
alnoki, line 87: I suggest you actually just link to the docs site here

In case we update the trade script, we don't want to have out-of-date terminal printouts

This introduces duplicate maintainer load

alnoki (changes requested):
See comments

src/docker/README.md
elliottdehn: Those instructions assume the user has to deploy the contracts and run the node which isn't the case here. I'll simply change this to specify what value for which prompt instead of having the printouts.

src/docker/README.md
```diff
@@ -35,15 +35,81 @@ docker compose --file src/docker/compose.dss-local.yaml down
...
+```
+
+Next, navigate your terminal to `/src/python/sdk` and run the following:
+
+```sh
+poetry install
+poetry run trade
 ```
 
+The script will have a few prompts, respond to each of them as follows:
```
alnoki, line 94: I'm able to render the markdown file locally, but to make sure that this runs on all platforms we probably want the blank link between this and the table that will get inserted if you run mdformat

alnoki (changes requested):

alnoki (approved):

PR #473, opened 2023-09-13
[ECO-481] Remove processor from cargo workspace

CRBl69:
This PR is intended to remove the processor from the cargo workspace to fix CI. It also reviled another problem with CI (an argument to a command that was copy pasted where it shouldn't have been) which was fixed.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -49,13 +49,13 @@ jobs:
         name: Install cargo-sort
         with:
           command: install
-          args: cargo-sort --manifest-path src/rust/Cargo.toml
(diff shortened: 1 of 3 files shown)


elliottdehn (approved):

elliottdehn:
When I do this on my machine, it takes away red underlines and hints in the processor project. I consider them necessary for development in Rust.

elliottdehn (changes requested):
I'm not sure what to do but it isn't an option to go without compiler hints and errors in IDE for me. Perhaps there's some other approach to address the issues?

elliottdehn (approved):

PR #474, opened 2023-09-19
[ECO-445] Add fill events

elliottdehn:
I've added fill events and can see them via postgrest at localhost:3000/fill_events.

For some reason the SQL wasn't working if I did it as a migration so I put it in the first one just for now. When Bogdan merges his changes to this SQL I'll use his pattern for creating accessibility to `web_anon`.

Remaining events are: **recognized market event**, **change order size event**, place market order event, and **place swap event**. Bolded are the events that the Python trade.py script is not currently triggering.

To add events:
0. Use CloudFlare VPN because devnet rate limits are trash.
1. Trigger the desired event(s) in the trade.py script so they can be easily re-used in the future for testing.
  - Note that this is also useful for other people who will be integrating Econia to be able to easily simulate all events.
  - You may instead decide to use the rust script for this, feel free.
2. Deploy a faucet and exchange to devnet using the `trade.py` instructions
3. Set your `econia_address` in your processor `config.yaml` which lives in /docker/processor to the address from above.
4. Run the script through to completion and take note of the first transaction hash: search for it [here](https://explorer.aptoslabs.com/transactions?network=devnet&type=user) and take note of the version number.
5. Set your `starting_version` in your processor `config.yaml` to the version number of the transaction above.
6. Add your database tables, data models and processor code to intake the new events into the database!
7. Run the processor using the docker compose file inside /src/docker! Remember: `config.yaml` goes inside /src/docker/processor.
  - Get an API key to access the devnet transaction stream [here](https://aptos-api-gateway-prod.firebaseapp.com/api-keys).
  - Here's an example complete `config.yaml`
 ```
health_check_port: 8085
server_config:
  processor_name: econia_processor
  indexer_grpc_data_service_address: grpc.devnet.aptoslabs.com:443
  auth_token: aptoslabs_W2Nc7fPfk3M_HHibCoJse65dWXjpS688ncusUzKXXXXXX
  econia_address: 0x9c6153cb50b815447fd73817547cbf71bcdeb26a33118c6c22f5b2c4e90c92b5
  starting_version: 2895526
```
8. Check that you can see the events at the views that you created, such as http://localhost:3001/market_registration_events
9. You're all done!

### CODE REVIEW

File: src/rust/dbv2/migrations/2023-08-25-111710_add_market_registration_events/up.sql [hunk 1]
```diff
@@ -43,6 +43,39 @@ AFTER INSERT ON market_registration_events FOR EACH ROW
...
+   PERFORM pg_notify('fill_event'::text, row_to_json(NEW)::text);
+   RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER fill_events_trigger
+AFTER INSERT ON fill_events FOR EACH ROW
+EXECUTE PROCEDURE notify_fill_event ();
+
```
CRBl69, line 78: I would have done this in a new migration. This one is called `market_registration_events`.

elliottdehn: That's true, I was having trouble doing it in a second migration so did it here. Can try again.

src/rust/dbv2/migrations/2023-08-25-111710_add_market_registration_events/down.sql
```diff
@@ -17,6 +17,24 @@ DROP FUNCTION notify_fill_event ();
...
+DROP TRIGGER cancel_order_events_trigger ON cancel_order_events;
+
+
+DROP FUNCTION notify_cancel_order_event ();
+
+
+DROP TABLE cancel_order_events;
+
+
 DROP TABLE IF EXISTS processor_status;
```
CRBl69, line 38: Dropping non existent objects here.

CRBl69 (changes requested):

CRBl69 (approved):

PR #479, opened 2023-09-21
Create missing DB types for events

CRBl69:


src/rust/dbv2/src/schema.rs
```diff
@@ -1,24 +1,640 @@
...
+        subdomain -> Varchar,
+        #[max_length = 66]
+        registered_address -> Nullable<Varchar>,
+        expiration_timestamp -> Nullable<Timestamp>,
+        #[max_length = 140]
+        token_name -> Varchar,
+        is_deleted -> Bool,
+        inserted_at -> Timestamp,
+    }
+}
```
alnoki, line 28: A lot of these tables are irrelevant for us, for example Aptos Name Service (ANS) tables

If these are included in here due to deisel autogen, is there anyway to exclude them?

alnoki (changes requested):
The migrations look good to me, but the schema includes a bunch of tables that are not relevant for our data service stack

alnoki (approved):

alnoki:
@CRBl69 Note that there are merge conflicts

PR #485, opened 2023-09-25
[ECO-483] Add order history !!!!!

CRBl69:
- Add database only docker file to ease development process.
- Add `aggregated_events` table to keep up which events were aggregated.
- Added 4 tables for user history:
    - `user_history` :point_right: general data.
    - `user_history_limit` :point_right: limit order specific data.
    - `user_history_market` :point_right:  market order specific data.
    - `user_hisotry_swap` :point_right: swap order specific data.
- Add user history processor to aggregator.
    - On each tick, takes into account all events that were not already aggregated (place, change, cancel, fill) and creates/updates the order.
- Fix naming in `add_limit_order_events/up.sql`.
- Add triggers in `event_types/up.sql`.


src/rust/dbv2/migrations/2023-09-20-065710_event_types/up.sql
```diff
@@ -0,0 +1,89 @@
...
+BEGIN
+   PERFORM pg_notify('change_order_size_event'::text, row_to_json(NEW)::text);
+   RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE TRIGGER change_order_size_events_trigger
+AFTER INSERT ON change_order_size_events FOR EACH ROW
+EXECUTE PROCEDURE notify_change_order_size_event ();
```
elliottdehn, line 27: Each of these needs something to this effect after the CREATE TRIGGER:

```
CREATE VIEW api.fill_events AS SELECT * FROM fill_events;

GRANT SELECT ON api.fill_events TO web_anon;
```

So that `web_anon` can see the events.

src/rust/dbv2/src/models.rs
```diff
@@ -50,16 +50,16 @@ pub struct PlaceLimitOrderEvent {
     pub event_idx: BigDecimal,
     pub time: DateTime<Utc>,
     pub market_id: BigDecimal,
-    pub maker_address: String,
-    pub maker_custodian_id: BigDecimal,
-    pub maker_order_id: BigDecimal,
-    pub maker_side: bool,
-    pub integrator_address: String,
+    pub user: String,
```
elliottdehn, line 53: You're going to need to update the processor to accommodate the new names.

src/rust/dbv2/migrations/2023-09-20-193413_add_limit_order_events/up.sql
```diff
@@ -43,16 +43,16 @@ CREATE TABLE
...
+    custodian_id NUMERIC(20) NOT NULL,
+    order_id NUMERIC(39) NOT NULL,
+    side BOOLEAN NOT NULL,
+    integrator VARCHAR(70) NOT NULL,
     initial_size NUMERIC(20) NOT NULL,
     price NUMERIC(20) NOT NULL,
-    restriction NUMERIC(20) NOT NULL,
-    self_match_behavior NUMERIC(20) NOT NULL,
-    posted_size NUMERIC(20) NOT NULL
+    restriction SMALLINT NOT NULL,
```
elliottdehn, line 53: You'll have to update the processor to accommodate the different type.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -0,0 +1,411 @@
...
+        self.last_indexed_timestamp.is_none()
+            || self.last_indexed_timestamp.unwrap() + Duration::seconds(5) < Utc::now()
+    }
+
+    async fn process_and_save_historical_data(&mut self) -> DataAggregationResult {
+        self.process_and_save_internal().await
+    }
+
+    fn poll_interval(&self) -> Option<std::time::Duration> {
+        Some(std::time::Duration::from_secs(60 * 60))
```
elliottdehn, line 50: I feel like we're going to want a much lower update interval for this, an hour latency is kind of high. I was thinking 1-5 seconds, it's not like this is a huge amount of work that it's doing.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -0,0 +1,411 @@
...
+            )
+            .execute(&self.pool)
+            .await
+            .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+        }
+        if !limit_events.is_empty() {
+            for x in limit_events {
+                sqlx::query!(
+                    r#"
+                        INSERT INTO aggregator.aggregated_events VALUES (
```
elliottdehn, line 174: Something tells me we should be using a SQL transaction here, because we might insert half the events to user history before erroring out. Then it won't insert any events into this table and we'll get duplicated work.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -0,0 +1,411 @@
...
+                    SELECT * FROM aggregator.aggregated_events
+                    WHERE fill_events.txn_version = aggregated_events.txn_version
+                    AND fill_events.event_idx = aggregated_events.event_idx
+                )
+            "#,
+        )
+        .fetch_all(&self.pool)
+        .await
+        .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+        let change_events = sqlx::query!(
```
elliottdehn, line 67: need to sort by the version and event idx so that we always aggregate the change size events in the order that they happen.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -0,0 +1,411 @@
...
+                    "#,
+                    x.txn_version,
+                    x.event_idx,
+                )
+                .execute(&self.pool)
+                .await
+                .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+            }
+        }
+        for x in &change_events {
```
elliottdehn, line 299: I suspect that with this you may get a size underflow if an order is created at 30, then changes size to 50, then fills 45, then changes size to 60, then fills 60. The total filled size is 105 but the size you'll have in the user history before you process fill events is 60.

I believe that you have to process/aggregate the events in the order that they happen. The txn version and event idx form a total ordering so that should help you.

elliottdehn (changes requested):
Need to update the processor, need to aggregate with a SQL transaction, and need to aggregate in the order of events as they occur. See comments :-)

src/rust/dbv2/migrations/2023-09-20-065710_event_types/up.sql
CRBl69: Done

src/rust/dbv2/src/models.rs
CRBl69: Yes, that will come in a pr on the other repo.

src/rust/dbv2/migrations/2023-09-20-193413_add_limit_order_events/up.sql
CRBl69: Yes, that will come in a pr on the other repo.

src/rust/aggregator/src/data/user_history.rs
CRBl69: Yes, I intended it to be 5 seconds as that's what I did in the `ready` function but forgot to do it here. Thanks for spotting that

src/rust/aggregator/src/data/user_history.rs
CRBl69: Yes, just did that :+1:

src/rust/aggregator/src/data/user_history.rs
CRBl69: The only case I could come up with where this could be an issue is if someone can close an order by changing size to 0. Is that a possibility ?

src/rust/aggregator/src/data/user_history.rs
CRBl69: The underflow isn't a problem IMO. If the events are indexed in order, that means the underflow will only last inside the transaction, but will be gone once the transaction is done (DB transactions). If they are not, then ordering or not ordering will not make a difference because, but still, the underflow might last max 5 secs between aggregator scans. We can talk more about this on a call.

elliottdehn (approved):

PR #489, opened 2023-09-27
close button border blue

chiumax:
modals with close button on hover will turn blue:
image

### DIFF
--- a/src/typescript/frontend/src/components/modals/BaseModal.tsx
+++ b/src/typescript/frontend/src/components/modals/BaseModal.tsx
@@ -23,6 +23,8 @@ export const BaseModal: React.FC<
   children,
   className,
 }) => {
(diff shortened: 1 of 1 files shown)


src/typescript/frontend/src/components/modals/BaseModal.tsx
```diff
@@ -23,6 +23,8 @@ export const BaseModal: React.FC<
   children,
   className,
 }) => {
+  const [hoveringOnCloseButton, setHoveringOnCloseButton] =
```
sotatek-stephen-pham, line 26: May I ask why we have to use the useState hook to store hovering state?

chiumax: because i'm unable to produce the same effect using pure css, so we need to track hover state to trigger blue outline

sotatek-stephen-pham (approved):

PR #495, opened 2023-09-28
[ECO-534] Aggregate events in a certain order

CRBl69:
Update the order in which events are aggregated in order to generate a correct representation of order history.


src/rust/aggregator/src/data/user_history.rs
```diff
@@ -285,51 +265,32 @@ impl Data for UserHistory {
...
-                )
-                .execute(&mut transaction as &mut PgConnection)
-                .await
-                .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-            }
+        for i in 0..fill_events.len().max(change_events.len()) {
+            match (fill_events.get(i), change_events.get(i)) {
+                (Some(fill), Some(change)) => {
+                    if fill.txn_version < change.txn_version || (fill.txn_version == change.txn_version && fill.event_idx < change.event_idx) {
+                        aggregate_fill(&mut transaction, &fill.size, &fill.maker_order_id, &fill.market_id, &fill.time).await?;
```
elliottdehn, line 275: This is incorrect. You might aggregate say, 8 fill events before a change size event. Just because the change size event is adjacent to the fill event doesn't mean that it gets aggregated immediately afterward.

alnoki (changes requested):
Will you please update this PR with a description?

src/rust/aggregator/src/data/user_history.rs
CRBl69: Done

alnoki (approved):

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -348,70 +357,94 @@ impl Data for UserHistory {
...
-                            $1, $2
-                        );
-                    "#,
-                    x.txn_version,
-                    x.event_idx,
-                )
-                .execute(&mut transaction as &mut PgConnection)
-                .await
-                .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+                mark_as_aggregated(&mut transaction, &x.txn_version, &x.event_idx).await?;
```
elliottdehn, line 360: It doesn't seem like we actually do anything with the cancel events other than marking them as aggregated. Shouldn't we be updating the status of the order to `cancelled`?

CRBl69: That is done in the loop above. Actually, let me merge the loops as it is useless and confusing that there are two of them.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -348,70 +357,94 @@ impl Data for UserHistory {
...
+        order_id,
+        market_id,
+        time,
+    )
+    .execute(tx as &mut PgConnection)
+    .await
+    .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+    Ok(())
+}
+
```
elliottdehn, line 431: suggest to add `aggregate_cancel` so that the status will be correct.

CRBl69:
```rust
for x in &cancel_events {
    sqlx::query!(
        r#"
            UPDATE aggregator.user_history
            SET order_status = 'cancelled', last_updated_at = $3
            WHERE order_id = $1 AND market_id = $2;
        "#,
        x.order_id,
        x.market_id,
        x.time,
    )
    .execute(&mut transaction as &mut PgConnection)
    .await
}
if !cancel_events.is_empty() {
    for x in cancel_events {
        mark_as_aggregated(&mut transaction, &x.txn_version, &x.event_idx).await?;
    }
}
```

It is done just above the `mark_as_aggregate` part. This is not a function as it is used only once (in contrast with fills and changes).

elliottdehn (approved):

PR #497, opened 2023-09-28
[ECO-521] Add e2e framework

CRBl69:
Add an e2e framework.

There are 3 crates added in this PR:

| Name  | Description |
| --- | --- |
| `e2e` | The main crate. Contains the different tests as well as the main function that runs all those tests.|
| `e2e-proc-macro` | A crate that contains the `e2e_test` proc macro. This macro transforms the return type of a function by wrapping it into `Metadata`. It can also accept a namespace as an argument (this is used to group tests into named spaces). |
| `metadata` | A crate containing a simple `Metadata` struct. This contains the name of the function, as well as the namespace of the test. |

To create a new test, you can look at `e2e/src/market_registration.rs` for inspiration.

### DIFF
--- a/src/rust/e2e/src/main.rs
+++ b/src/rust/e2e/src/main.rs
@@ -1,177 +1,84 @@
-use std::{str::FromStr, time::Duration};
+use std::{collections::HashMap, future::Future, pin::Pin};
 
-use anyhow::{ensure, Result};
-use aptos_sdk::{
-    bcs,
-    move_types::{
-        ident_str,
-        language_storage::{ModuleId, StructTag, TypeTag},
-    },
-    rest_client::{aptos_api_types::MoveModuleId, FaucetClient},
-    types::{
-        account_address::AccountAddress, transaction::EntryFunction, LocalAccount, APTOS_COIN_TYPE,
-    },
-};
+use anyhow::Result;
 use clap::Parser;
-use econia_sdk::{entry::*, errors::EconiaError, EconiaClient, EconiaResult};
-use sqlx::{query, PgPool};
+use colored::Colorize;
+use futures::{stream::FuturesUnordered, StreamExt};
+use indicatif::ProgressStyle;
+use market_registration::test_market_registration;
+use metadata::Metadata;
+use utils::{init, Args, State};
 
-const TIMEOUT: Duration = Duration::from_secs(1);
+mod market_registration;
+mod utils;
 
-#[derive(Parser, Debug)]
-pub struct Args {
-    /// The URL of the Aptos node
-    pub node_url: String,
-
-    /// The URL of the faucet
-    pub faucet_url: String,
-
-    /// The address of the Econia contract
-    pub econia_address: String,
-
-    /// The address of the Aptos faucet
-    pub faucet_address: String,
-
-    /// The database URL
-    pub db_url: String,
+fn message(success: bool, name: &str) -> String {
+    let status = if success { "OK".green() } else { "FAIL".red() };
+    let name = if name.len() > 20 {
+        format!("{}...", &name[..17])
+    } else {
+        name.to_string()
+    };
+    format!("{name}: {status}")
 }
 
-pub struct Init {
-    e_apt: TypeTag,
-    e_usdc: TypeTag,
-    faucet_address: AccountAddress,
-    faucet_client: FaucetClient,
-    econia_address: AccountAddress,
-    econia_client: EconiaClient,
-    db_pool: PgPool,
-}
+async fn run_tests(state: State) -> Result<()> {
+    let mut successes: HashMap<String, Vec<String>> = HashMap::new();
+    let mut fails: HashMap<String, Vec<(String, anyhow::Error)>> = HashMap::new();
+    {
+        let mut v: FuturesUnordered<Pin<Box<dyn Future<Output = Metadata<Result<()>>>>>> =
+            FuturesUnordered::new();
+        v.push(Box::pin(test_market_registration(&state)));
 
-/// Creates the initial variables needed
-async fn init(args: &Args) -> Init {
-    // Create eAPT and eUSDC `TypeTag`s
-    let faucet_address = AccountAddress::from_hex_literal(&args.faucet_address).unwrap();
-    let e_apt = TypeTag::Struct(Box::new(
-        StructTag::from_str(&format!("0x{faucet_address}::example_apt::ExampleAPT")).unwrap(),
-    ));
-    let e_usdc = TypeTag::Struct(Box::new(
-        StructTag::from_str(&format!("0x{faucet_address}::example_usdc::ExampleUSDC")).unwrap(),
-    ));
-
-    // Create a `FaucetClient`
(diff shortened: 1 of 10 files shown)


src/rust/e2e/src/market_registration.rs
```diff
@@ -0,0 +1,45 @@
...
+    .unwrap();
+
+    let (_, mut econia_client) =
+        account(&state.faucet_client, &state.node_url, state.econia_address).await;
+
+    econia_client.submit_tx(entry).await?;
+
+    std::thread::sleep(TIMEOUT);
+
+    let result = query!("SELECT COUNT(*) AS count FROM market_registration_events")
```
elliottdehn, line 35: Isn't there a delay interval on all the aggregations, so this might (will probably) spuriously fail?
Probably need to be mindful of the delays and wait twice as long to give the aggregator time.

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub db_pool: PgPool,
+    pub node_url: String,
+}
+
+/// Creates the initial variables needed
+pub async fn init(args: &Args) -> State {
+    // Create eAPT and eUSDC `TypeTag`s
+    let faucet_address = AccountAddress::from_hex_literal(&args.faucet_address).unwrap();
+    let e_apt = TypeTag::Struct(Box::new(
+        StructTag::from_str(&format!("0x{faucet_address}::example_apt::ExampleAPT")).unwrap(),
```
elliottdehn, line 55: The address format that I've seen everywhere is to accept 0x prefixes and "fix" non-prefixed addresses silently.

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,133 @@
...
+        .await
+        .expect("Could not connect to the database.");
+
+    State {
+        e_apt,
+        e_usdc,
+        faucet_address,
+        faucet_client,
+        econia_address,
+        econia_client,
```
elliottdehn, line 83: We're probably going to want two clients, so that we can simulate accounts trading against each other.

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,133 @@
...
+        None,
+    )
+    .await
+    .unwrap();
+
+    (account_address, econia_client)
+}
+
+/// Funds an amount with the coin specified
+pub async fn fund(
```
elliottdehn, line 116: I think you forgot to use this to fund the account(s) with eAPT and eUSDC. I'd leave it to the tests to deposit since we're going to need to write a test for account balances.

One further thought, how would we write such a test? All of the tests will run at the same time and deposit different amounts, so how can we ensure that the account balance test runs as expected?

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,133 @@
...
+    let econia_address =
+        AccountAddress::from_hex_literal(&args.econia_address).expect("Could not parse address.");
+
+    let (_, econia_client) = account(&faucet_client, &args.node_url, econia_address.clone()).await;
+
+    let db_pool = PgPool::connect(&args.db_url)
+        .await
+        .expect("Could not connect to the database.");
+
+    State {
```
elliottdehn, line 77: State will probably need a thread-safe integer for the min size of the market, which gets accessed and increased with every test. This way each test can create its own market safely.

src/rust/e2e/src/utils.rs
alnoki: I would think we can use the same client just with separate asyncronous account txns for this, correct?

src/rust/e2e/src/main.rs
```diff
@@ -1,177 +1,84 @@
...
 
-    let econia_client = EconiaClient::connect(
-        reqwest::Url::parse(&node_url).unwrap(),
-        econia_address.clone(),
-        account,
-        None,
-    )
-    .await
-    .unwrap();
+    println!("Sccesses:");
```
alnoki, line 62: Typo "Successes"

alnoki (changes requested):
How about?

1. A description on the PR
2. A README that shows how to run the tests, referencing the local DSS, etc.

alnoki (commented):
How about?

1. A description on the PR
2. A README that shows how to run the tests, referencing the local DSS, etc.

src/rust/e2e/src/utils.rs
CRBl69: You can always create more clients. There are just some provided.

src/rust/e2e/src/utils.rs
CRBl69: They use new accounts.

src/rust/e2e/src/market_registration.rs
CRBl69: I would suggest having a very small timeout on the aggregator while testing when possible.

src/rust/e2e/src/utils.rs
alnoki: I consider this resolved, standing by for @elliottdehn to mark conversation as resolved

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+    // Create eAPT and eUSDC `TypeTag`s
+    let faucet_address = if args.faucet_address.starts_with("0x") {
+        args.faucet_address.clone()
+    } else {
+        format!("0x{}", args.faucet_address)
+    };
+    let faucet_address =
+        AccountAddress::from_hex_literal(&faucet_address).expect("Could not parse faucet address.");
+    let e_apt = TypeTag::Struct(Box::new(
+        StructTag::from_str(&format!("0x{faucet_address}::example_apt::ExampleAPT")).unwrap(),
```
elliottdehn, line 60: It either already starts with 0x, or gets prefixed with 0x, so you can get rid of the 0x prefix here I think

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+    ));
+    let e_usdc = TypeTag::Struct(Box::new(
+        StructTag::from_str(&format!("0x{faucet_address}::example_usdc::ExampleUSDC")).unwrap(),
```
elliottdehn, line 63: same down here

src/rust/e2e/src/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+    State {
+        e_apt,
+        e_usdc,
+        faucet_address,
+        faucet_client,
+        econia_address,
+        econia_client,
+        db_pool,
+        node_url: args.node_url.clone(),
+        market_size: AtomicU64::new(1),
```
elliottdehn, line 96: Hmmm I think you'll need to check the number of registered markets and start there (plus one). This makes it so that you can only run the tests once for a given chain instance, which I imagine we'll be running them multiple times as we change things.

CRBl69: I added logic to query the min size on start.

alnoki (approved):

PR #504, opened 2023-09-29
[ECO-562] Evict empty orders during matching

alnoki:
The matching engine returns silently if it encounters an order of size 0, which should never happen in production.

Add an extra check to evict an order of size 0 if one ends up on the book, rather than returning silently. Also add an associated test.

### DIFF
--- a/src/move/econia/sources/market.move
+++ b/src/move/econia/sources/market.move
@@ -2983,6 +2983,7 @@ module econia::market {
     /// * `test_match_empty()`
     /// * `test_match_fill_size_0()`
     /// * `test_match_loop_twice()`
+    /// * `test_match_order_size_0()`
     /// * `test_match_partial_fill_lot_limited_sell()`
     /// * `test_match_partial_fill_tick_limited_buy()`
     /// * `test_match_price_break_buy()`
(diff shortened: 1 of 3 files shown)


src/move/econia/sources/market.move
```diff
@@ -3077,6 +3078,31 @@ module econia::market {
...
+            // verify the assumption of no empty orders on the book.
+            if (order_ref_mut.size == 0) {
+                let Order{
+                    size: evictee_size,
+                    price: evictee_price,
+                    user: evictee_user,
+                    custodian_id: evictee_custodian_id,
+                    order_access_key: evictee_order_access_key
+                } = avl_queue::pop_head(orders_ref_mut);
+                user::cancel_order_internal(
```
elliottdehn, line 3093: Should we emit a fill event? There was no amount filled but we did match and remove an order from the book, which might be important to off-chain API's that are expecting either a fill or a cancel to remove the order from the book.

src/move/econia/sources/market.move
```diff
@@ -3077,6 +3078,31 @@ module econia::market {
...
+                    market_id,
+                    evictee_custodian_id,
+                    side,
+                    evictee_size,
+                    evictee_price,
+                    evictee_order_access_key,
+                    (NIL as u128),
+                    CANCEL_REASON_EVICTION,
+                );
+                continue
```
elliottdehn, line 3104: Should we emit a fill event? There was no amount filled but we did match and remove an order from the book, which might be important to off-chain API's that are expecting either a fill or a cancel to remove the order from the book.

elliottdehn (approved):

PR #512, opened 2023-10-02
[ECO-575] Fix base type processing (submodule update)

elliottdehn:
See https://github.com/econia-labs/aptos-indexer-processors/pull/10


alnoki (changes requested):
`aptos-core` is introduced as a submodule but isn't needed

elliottdehn:
Fixed

alnoki (approved):

PR #513, opened 2023-10-02
[ECO-581][ECO-580] Create script to verify WebSocket functionality & switch WS/REST ports

elliottdehn:
Due to how the database OpenAPI is autogenerated, it includes a port `3000` which is the internal port for the docker instance. We're using a `3001` external port, which means the webpage can't be used as a REST client for a local instance. It's easier to switch the ports and update documentation than it necessarily is do change it manually in the docs every time or in CI. I've done that here, and verified that WebSockets continue to work with a small example script that can be used to listen to any single channel and print out the messages. This is useful for validation and for users who may find such an example helpful.


src/docker/compose.dss-local.yaml
```diff
@@ -42,6 +42,7 @@ services:
       - HEALTHCHECK_BEFORE_START=true
 
   ws:
+    platform: linux/x86_64
```
alnoki, line 45: Ditto here for the platform issue

alnoki (changes requested):
I had a few suggestions for layout/walkthrough content that were more easily captured as a PR: https://github.com/econia-labs/econia/pull/514

I don't think those changes should be contentious and so I suggest we merge them in while identifying a solution to the outstanding platform issue I discuss in my comments

src/docker/compose.dss-global.yaml
```diff
@@ -31,6 +31,7 @@ services:
       - HEALTHCHECK_BEFORE_START=false
 
   ws:
+    platform: linux/x86_64
```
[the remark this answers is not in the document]
elliottdehn: You're correct that it shows a warning on Mac. However, if I don't include that, then the container exits on starting with an error: https://stackoverflow.com/questions/68630526/lib64-ld-linux-x86-64-so-2-no-such-file-or-directory-error

I don't _think_ it should make compiling impossible in some cases. Can you explain how you came to that conclusion?
Not sure what the solution to this is if not that. @CRBl69 had no issue because he's on Linux already.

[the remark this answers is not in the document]
alnoki: Noting here as discussed offline that the issue could be solved by compiling postgres-websockets from source in `Dockerfile.ws`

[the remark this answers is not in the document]
alnoki: Tagging https://github.com/diogob/postgres-websockets/issues/93

src/docker/api/Dockerfile.ws
```diff
@@ -1,13 +1,24 @@
...
-RUN chmod u+x ./postgres-websockets
-
-ENTRYPOINT ["./postgres-websockets"]
+WORKDIR /root/.local/bin
+ENV PATH="/root/.local/bin:$PATH"
+WORKDIR /app
+RUN wget -qO- https://get.haskellstack.org/ | sh
+RUN git clone https://github.com/diogob/postgres-websockets.git
+WORKDIR /app/postgres-websockets
+RUN echo $HOME
```
alnoki, line 20: This can be useful for diagnostics during development, but let's take this out once there is a stable solution

src/docker/api/Dockerfile.ws
```diff
@@ -1,13 +1,24 @@
...
+        zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
 
-RUN wget https://github.com/diogob/postgres-websockets/releases/download/$POSTGRES_WEBSOCKETS_VERSION/postgres-websockets
-
-RUN chmod u+x ./postgres-websockets
-
-ENTRYPOINT ["./postgres-websockets"]
+WORKDIR /root/.local/bin
+ENV PATH="/root/.local/bin:$PATH"
```
alnoki, line 15: Is this necessary if we are able to resolve the `stack build ` vs `stack install` issue?

src/docker/api/Dockerfile.ws
```diff
@@ -1,13 +1,24 @@
...
+WORKDIR /app
+RUN wget -qO- https://get.haskellstack.org/ | sh
+RUN git clone https://github.com/diogob/postgres-websockets.git
+WORKDIR /app/postgres-websockets
+RUN echo $HOME
+RUN stack setup
+RUN stack build
+RUN cd .stack-work/install/*/*/*/bin && mv * /root/.local/bin
```
alnoki, line 23: I understand we should be able to fix this via `stack install` per https://github.com/diogob/postgres-websockets/issues/93#issuecomment-1749672418

alnoki: Yep it looks like the README is updated, can you simplify this to follow?

https://github.com/diogob/postgres-websockets#building-from-source

src/docker/api/Dockerfile.ws
```diff
@@ -1,13 +1,24 @@
 FROM debian:stable-slim
```
alnoki, line 1: Use https://hub.docker.com/_/haskell/, which will have the official stack toolchain already installed

alnoki (changes requested):
I'm already signed off on everything except for the websockets Dockerfile

Unlike most of our Dockerfiles which use Rust-based projects, this one uses [`postgres-websockets`](https://github.com/diogob/postgres-websockets), which is implemented in Haskell.

To compile from source it suggests we use [Stack](https://docs.haskellstack.org/en/stable/), which is analogous to the Rust toolchain. Stack comes preloaded on the [official Haskell Docker image](https://hub.docker.com/_/haskell/), so lets use that instead of `debian:stable-slim`

Basically the way we are doing it now is analogous to using a base Linux image, installing Rustup, installing Rust, then compiling a Rust project from source, rather than using a Rust image and compiling from source.

Note too that the [`postgres-websockets`](https://github.com/diogob/postgres-websockets) README has been updated to reflect the issue uncovered in https://github.com/diogob/postgres-websockets/issues/93#issuecomment-1749672418 , so we should be able to compile from source much more simply

My main concern is that it will work on practically every platform we need it to, especially GCP, and so we should use the most official toolchain steps as possible, which involves the official Docker Haskell image and following the (updated) `postgres-websockets` install verbatim

src/docker/api/Dockerfile.ws
elliottdehn: Yes, we do have to add it to our PATH

elliottdehn:
I was able to fix the executable hack using the approach suggested by the creator but failed to use a haskell image per a comment I made here: https://github.com/diogob/postgres-websockets/issues/93#issuecomment-1751324571

alnoki (approved):

PR #518, opened 2023-10-04
[ECO-484] Add order book state

CRBl69:
- add a prev column to limit orders
- updated views (when updating the underlying tables, views need
  to be updated)
- fix bigdecimal dependency
- fix sdk (broke by aptos update)
- fix fills (orders would not go into 'closed' mode)

Order is implemented with a field named `last_increase_stamp`. This is a number where the order with the lowest `last_increase_stamp` is the highest priority one. It is computed as `(txn_version << 64) | event_idx`.

The `limit_orders` endpoint is basically the order book state endpoint.

Example query: `http://0.0.0.0:3001/limit_orders?order=price.desc,last_increase_stamp.asc&market_id=eq.3&side=eq.ask&order_status=eq.open&limit=3

Example result:

```json
[
  {
    "market_id": 3,
    "order_id": 866997112223312137760,
    "user": "0x9538c839fe490ccfaf32ad9f7491b5e84e610ff6edc110ff883f06ebde82463d",
    "custodian_id": 0,
    "self_matching_behavior": 3,
    "restriction": 0,
    "created_at": "2023-09-14T04:39:15.732539+00:00",
    "last_updated_at": null,
    "integrator": "0xd718181a753f5b759518d9b896018dd7eb3d77d80bf90ba77fffaf678f781929",
    "total_filled": 0,
    "remaining_size": 10000,
    "order_status": "open",
    "order_type": "limit",
    "price": 20000,
    "last_increase_stamp": 12466419366246665406505811969,
    "side": "ask"
  },
  {
    "market_id": 3,
    "order_id": 1272825481853512207904,
    "user": "0x9538c839fe490ccfaf32ad9f7491b5e84e610ff6edc110ff883f06ebde82463d",
    "custodian_id": 0,
    "self_matching_behavior": 3,
    "restriction": 0,
    "created_at": "2023-09-16T07:45:30.393582+00:00",
    "last_updated_at": null,
    "integrator": "0xd718181a753f5b759518d9b896018dd7eb3d77d80bf90ba77fffaf678f781929",
    "total_filled": 0,
    "remaining_size": 10000,
    "order_status": "open",
    "order_type": "limit",
    "price": 20000,
    "last_increase_stamp": 12514843600519911097390596097,
    "side": "ask"
  },
  {
    "market_id": 3,
    "order_id": 1291272225935811694112,
    "user": "0x9538c839fe490ccfaf32ad9f7491b5e84e610ff6edc110ff883f06ebde82463d",
    "custodian_id": 0,
    "self_matching_behavior": 3,
    "restriction": 0,
    "created_at": "2023-09-16T07:48:19.409595+00:00",
    "last_updated_at": null,
    "integrator": "0xd718181a753f5b759518d9b896018dd7eb3d77d80bf90ba77fffaf678f781929",
    "total_filled": 0,
    "remaining_size": 10000,
    "order_status": "open",
    "order_type": "limit",
    "price": 20000,
    "last_increase_stamp": 12514871842485087946714120193,
    "side": "ask"
  }
]
```

Config used:

```yaml
health_check_port: 8085
server_config:
  processor_config:
    type: econia_transaction_processor
    econia_address: 0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff
  indexer_grpc_data_service_address: https://grpc.testnet.aptoslabs.com:443
  auth_token: TOKEN
  starting_version: 649555969
```


src/rust/aggregator/src/main.rs
```diff
@@ -22,9 +22,9 @@ async fn main() -> Result<()> {
 
     let mut data: Vec<Arc<Mutex<dyn Data + Send + Sync>>> = vec![];
 
-    data.push(Arc::new(Mutex::new(MarketsRegisteredPerDay::new(
-        pool.clone(),
-    ))));
+    // data.push(Arc::new(Mutex::new(MarketsRegisteredPerDay::new(
```
elliottdehn, line 25: Why is this commented out, and if so, why isn't it deleted instead?
Commented out code is a smell

alnoki (changes requested):
Additional TODOs that I could not attach directly to comments

1. Take out the `aptos-indexer-processors` submodule update
2. When ready to merge and all comments addressed, add in the description a printout like the one in #511
  a. Show the config used for global testing and some of the printouts (e.g. screenshots like in https://github.com/econia-labs/aptos-indexer-processors/pull/9#discussion_r1343233235)
  b. Potentially use a more recent version number to hit relevant functionality
  c. Also run the local DSS per the docker readme to verify it works, since that docker compose has changed

src/rust/dbv2/migrations/2023-10-03-071834_fix_price_field/down.sql
```diff
@@ -0,0 +1,7 @@
+-- This file should undo anything in `up.sql`
```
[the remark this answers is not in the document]
elliottdehn: Consolidating them requires you to drop the whole DB and set it back up with mock data every time you change the database. Having multiple migrations lets you continue to use the same mock data as you futz about with the codebase.

I can see why you might want them in one folder but there's a time-saving reason why someone might do it this way.

src/rust/dbv2/migrations/2023-10-03-071834_fix_price_field/up.sql
```diff
@@ -0,0 +1,9 @@
+-- Your SQL goes here
+DROP VIEW api.limit_orders;
+
+ALTER TABLE aggregator.user_history_limit ADD COLUMN price NUMERIC(20) NOT NULL;
```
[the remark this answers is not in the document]
elliottdehn: It says `user_history_limit` so I assume it's just limit orders?

src/rust/dbv2/migrations/2023-10-03-070930_add_order_order/up.sql
```diff
@@ -0,0 +1,25 @@
...
+DROP VIEW api.limit_orders;
+
+ALTER TABLE aggregator.user_history_limit ADD COLUMN prev NUMERIC;
+
+CREATE VIEW api.limit_orders AS
+    SELECT * FROM aggregator.user_history_limit NATURAL JOIN aggregator.user_history;
+
+GRANT SELECT ON api.limit_orders TO web_anon;
+
+CREATE FUNCTION linked_list_last(NUMERIC(20), NUMERIC(20)) RETURNS NUMERIC AS $$
```
elliottdehn, line 11: Do we necessarily need the linked list data structure in the database? I don't think so.
We should be able to select orders in the open state, ordered by price and THEN time, where market_id equals given. This is the limit order book state for the given market.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -70,7 +70,7 @@ impl Data for UserHistory {
         .fetch_all(&mut transaction as &mut PgConnection)
         .await
         .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-        let mut change_events = sqlx::query!(
+        let change_events = sqlx::query!(
```
elliottdehn, line 73: Change size increase sends the order to the back of its price time priority queue. Implement the book state as a view the way I suggested and you can account for this by updating the order's timestamp.

elliottdehn (changes requested):
Need to account for change size (increase) event, reconsider use of persistent data structure.

elliottdehn:
How does this handle bids/asks having separate books? I think the data structure assumes all markets have just one book which isn't true.

elliottdehn:
You should be able to provide two views: `order_book_asks` and `order_book_bids` which are simply a select of all orders where side is (given, constant) and market id is (given, optional) and status is OPEN, ordered by price and timestamp. Timestamp is updated on order change size increases. No need for the persistent linked list or any next/previous logic.

src/rust/aggregator/src/main.rs
CRBl69: Big smell indeed. I did this for testing, forgot to undo. Thanks for spotting.

src/docker/compose.dss.yaml
```diff
@@ -0,0 +1,44 @@
...
+      context: ../../
+      dockerfile: src/docker/api/Dockerfile.ws
+      args:
+        - POSTGRES_WEBSOCKETS_VERSION=0.11.1.0
+    depends_on:
+      - diesel
+      - postgres
+    environment:
+      - PGWS_DB_URI=postgres://econia:econia@postgres/econia
+      - PGWS_JWT_SECRET=econia_is_dooooooooooooooooooope
```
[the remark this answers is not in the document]
CRBl69: Haha. It has to be at least 32 chars long. So how about `conjunctivodacryocystorhinostomy`? It seems to be a 32-letter word according to [this source](https://en.wiktionary.org/wiki/conjunctivodacryocystorhinostomy).

src/rust/dbv2/migrations/2023-10-03-071834_fix_price_field/up.sql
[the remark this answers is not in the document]
CRBl69: Yes, these are just limit orders.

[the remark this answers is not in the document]
CRBl69: There is `user_history` which contains everything that is common to all types of orders (order id, market id, etc.), and then there's `user_history_{limit,market,swap}` which contains everything that is specific to each.

src/rust/dbv2/src/schema.rs
```diff
@@ -1,3 +1,22 @@
+// @generated automatically by Diesel CLI.
```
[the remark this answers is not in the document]
CRBl69: IMO this is due to the following order:

- Someone added `market_account_handles`.
- Someone added `balance_updates_by_handle`.
- Those got merged by git in the wrong order.
- `diesel` reordered them.

But I might be wrong. I don't know of any feature to reorder them.

src/rust/dbv2/migrations/2023-10-03-070930_add_order_order/up.sql
CRBl69: This would work if only we did not move the order at the end of the queue when changing size to a higher amount. But you give me an idea. What about adding a `last_time_size_went_up` field (TODO: find a better name), and sort by `COALESCE(last_time_size_went_up, creation_time)` (returns first if not null, second else). @alnoki would this work ? This would greatly simplify the logic and remove room for error.

elliottdehn: Since all transactions in the same block have the same timestamp, you have to use (transaction version, event index) instead of a traditional timestamp. This is because multiple orders could be created and resized inside of one transaction.

I guess we could include the creation "time" but I was thinking we could get by with just `last_increase_txn/idx` which is initialized to the txn/idx of the creation event and updated with change size increases.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -150,6 +150,8 @@ impl Data for UserHistory {
                 x.side,
                 x.self_match_behavior,
                 x.restriction,
+                x.price,
+                x.time,
```
elliottdehn, line 154: You need to use transaction version and event index instead of time because (as mentioned in Slack) one block has many transactions with the same time and one transaction may have many orders with the same version.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -396,6 +398,39 @@ async fn aggregate_change<'a>(
...
+    .await
+    .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+    let (order_type, original_size): (OrderType, BigDecimal) =
+        (record.order_type, record.remaining_size);
+    // If its a limit order and needs reordering
+    if matches!(order_type, OrderType::Limit) && &original_size < new_size {
+        sqlx::query!(
+            r#"
+                UPDATE aggregator.user_history_limit
+                SET last_increase_time = $3
```
elliottdehn, line 422: Here again, need to use transaction version and event index.

src/rust/dbv2/migrations/2023-10-03-070930_add_order_queue/up.sql
```diff
@@ -0,0 +1,130 @@
...
+      y api.market_orders%ROWTYPE;
+      z api.swap_orders%ROWTYPE;
+BEGIN
+   IF NEW.order_type = 'limit' THEN
+      SELECT * INTO x FROM api.limit_orders AS lo WHERE lo.market_id = NEW.market_id AND lo.order_id = NEW.order_id;
+      PERFORM pg_notify('econiaws', json_build_object('channel', 'updated_limit_order', 'payload', x)::text);
+   END IF;
+   IF NEW.order_type = 'market' THEN
+      SELECT * INTO y FROM api.market_orders AS lo WHERE lo.market_id = NEW.market_id AND lo.order_id = NEW.order_id;
+      PERFORM pg_notify('econiaws', json_build_object('channel', 'updated_market_order', 'payload', y)::text);
```
elliottdehn, line 117: I don't think that market orders and swap orders can be updated. Are those two removable? If so we should remove them.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -395,7 +404,45 @@ async fn aggregate_change<'a>(
...
+        (record.order_type, record.remaining_size);
+    // If its a limit order and needs reordering
+    if matches!(order_type, OrderType::Limit) && &original_size < new_size {
+        let txn = txn_version.to_bigint().ok_or(DataAggregationError::ProcessingError(anyhow!("txn_version not integer")))? << 64;
+        let event = event_idx.to_bigint().ok_or(DataAggregationError::ProcessingError(anyhow!("event_idx not integer")))?;
+        let txn_event: BigDecimal = BigDecimal::from(txn & event);
+        sqlx::query!(
+            r#"
+                UPDATE aggregator.user_history_limit
+                SET last_increase_stamp = $3
```
elliottdehn, line 434: Don't you need to use txn version and event idx here too?

CRBl69: image

It is using txn/idx. The thing is called stamp. Maybe I misunderstand the meaning of stamp but for me I replaced the time in timestamp with txn/idx and just called it stamp

elliottdehn (approved):

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -288,6 +293,8 @@ impl Data for UserHistory {
                             &change.order_id,
                             &change.market_id,
                             &change.time,
+                            &fill.txn_version,
+                            &fill.event_idx,
```
elliottdehn, line 297: @CRBl69 It doesn't look like you de-duplicate fill events in this code here. There two fill events emitted for every fill, one for the maker and one for the taker. You can de-duplicated them using the order id field (I believe).

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -137,10 +137,13 @@ impl Data for UserHistory {
         .await
         .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
         for x in &limit_events {
+            let txn = x.txn_version.to_bigint().ok_or(DataAggregationError::ProcessingError(anyhow!("txn_version not integer")))? << 64;
+            let event = x.event_idx.to_bigint().ok_or(DataAggregationError::ProcessingError(anyhow!("event_idx not integer")))?;
+            let txn_event: BigDecimal = BigDecimal::from(txn & event);
```
elliottdehn, line 142: @CRBl69 You mean `(txn << 64) | (event)`

elliottdehn (changes requested):

src/rust/aggregator/src/data/user_history.rs
CRBl69: The `<< 64` is done at line 140.

alnoki (approved):

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -300,14 +323,16 @@ impl Data for UserHistory {
...
                 (Some(fill), None) => {
-                    aggregate_fill(
-                        &mut transaction,
-                        &fill.size,
-                        &fill.maker_order_id,
-                        &fill.market_id,
-                        &fill.time,
-                    )
-                    .await?;
+                    if fill.maker_address == fill.emit_address {
```
alnoki, line 326: This check should go in `aggregate_fill`, because now market orders and swaps will not have their order size or status updated on fill

alnoki: Actually, deduping here works, but there needs to be extra logic in `aggregate_fill` that updates the maker and the taker side of the order

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -376,7 +403,7 @@ async fn aggregate_fill<'a>(
                     END
                 END,
                 last_updated_at = $4
-            WHERE order_id = $2 AND market_id = $3;
+            WHERE order_id = $2 AND market_id = $3
```
alnoki, line 406: With the recent deduping logic updates, I think this is off

Basically this function only gets called for the fill event emitted to the maker, but the changes need to be applied to the taker as well

So yes we should dedupe fill events by only calling this function for when the maker address is the address that the event was emitted to

But the taker *and* maker orders, defined by their order IDs as included in the fill event, need to be updated

alnoki (changes requested):

src/rust/aggregator/src/data/user_history.rs
alnoki: It also looks like this handles swaps but not market orders?

alnoki: I understand that anytime a fill event is applied to a swap the status is updated to closed, but this should only happen when remaining size is 0

alnoki: Fix in #525

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -59,6 +59,9 @@ impl Data for UserHistory {
             .begin()
             .await
             .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+        transaction.execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;")
```
alnoki, line 62: Noting here that for postgres this is effectively equal to `SERIALIZABLE`

https://www.postgresql.org/docs/current/transaction-iso.html

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -56,7 +59,10 @@ impl Data for UserHistory {
             .begin()
             .await
             .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-        let mut fill_events = sqlx::query!(
+        transaction.execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;")
+            .await
+            .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
```
alnoki, line 64: Actually, can we change this to `SERIALIZABLE`?

`REPEATABLE READ` has possible serialization anomolies in postgres: https://www.postgresql.org/docs/current/transaction-iso.html

And the aggregator is assumed to be effectively serial for this data stream

alnoki (changes requested):

alnoki (approved):

elliottdehn (approved):

PR #534, opened 2023-10-10
[ECO-661] Add logging to aggregator

elliottdehn:
This PR adds some logging to the aggregator to help diagnose crashing issues. It uses the `tracing` library which is used by the processor to do the same. I've primarily added logging to `UserHistory` but could add logging to the daily markets registered, although I thought that might be pointless spam.

---

Validated by observing these logs periodically in the docker output:
```
docker-aggregator-1  | 2023-10-10T20:25:22.468137Z  INFO aggregator: [Aggregator] Started up.
docker-aggregator-1  | 2023-10-10T20:25:22.481954Z  INFO aggregator: [Aggregator] Connected to DB.
docker-aggregator-1  | 2023-10-10T20:25:22.482049Z  INFO aggregator: [Aggregator] Starting process & save (historical, MarketsRegisteredPerDay).
docker-aggregator-1  | 2023-10-10T20:25:22.482217Z  INFO aggregator: [Aggregator] Starting process & save (historical, UserHistory).
docker-aggregator-1  | 2023-10-10T20:25:22.488307Z  INFO aggregator: [Aggregator] Finished process & save (historical, MarketsRegisteredPerDay).
docker-aggregator-1  | 2023-10-10T20:25:22.495964Z  INFO aggregator: [Aggregator] Finished process & save (historical, UserHistory).
docker-aggregator-1  | 2023-10-10T20:25:27.497253Z  INFO aggregator::data: [Aggregator] Start processing & saving (UserHistory)
docker-aggregator-1  | 2023-10-10T20:25:27.511502Z  INFO aggregator::data: [Aggregator] Finish processing & saving (UserHistory)
```

### DIFF
--- a/src/rust/aggregator/src/data/user_history.rs
+++ b/src/rust/aggregator/src/data/user_history.rs
@@ -40,16 +40,18 @@ impl UserHistory {
 
 #[async_trait::async_trait]
 impl Data for UserHistory {
+
+    fn model_name(&self) -> &'static str {
+        "UserHistory"
+    }
+
     fn ready(&self) -> bool {
         self.last_indexed_timestamp.is_none()
             || self.last_indexed_timestamp.unwrap() + Duration::seconds(5) < Utc::now()
     }
 
     async fn process_and_save_historical_data(&mut self) -> DataAggregationResult {
-        tracing::info!("[Aggregator] Start processing UserHistory.");
-        let res = self.process_and_save_internal().await;
-        tracing::info!("[Aggregator] Finish processing UserHistory.");
-        return res;
+        self.process_and_save_internal().await
     }
 
     fn poll_interval(&self) -> Option<std::time::Duration> {
@@ -60,7 +62,6 @@ impl Data for UserHistory {
     /// are also handled in a single atomic transaction for each batch of transactions, such that
     /// user history aggregation logic is effectively serialized across historical chain state.
     async fn process_and_save_internal(&mut self) -> DataAggregationResult {
-        tracing::info!("[Aggregator] Starting process & save (UserHistory).");
         let mut transaction = self
             .pool
             .begin()
@@ -69,7 +70,6 @@ impl Data for UserHistory {
         transaction.execute("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;")
             .await
             .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-
         let fill_events = sqlx::query!(
             r#"
                 SELECT * FROM fill_events
@@ -84,7 +84,6 @@ impl Data for UserHistory {
         .fetch_all(&mut transaction as &mut PgConnection)
         .await
         .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-
         let change_events = sqlx::query!(
             r#"
                 SELECT * FROM change_order_size_events
@@ -99,7 +98,6 @@ impl Data for UserHistory {
         .fetch_all(&mut transaction as &mut PgConnection)
         .await
         .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
-
         let cancel_events = sqlx::query!(
             r#"
                 SELECT * FROM cancel_order_events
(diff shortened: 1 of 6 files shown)


alnoki (changes requested):
Is there any way to abstract it and make it apply across all data models automatically?

For example if the aggregator panics due to an omitted endpoint then we will have no visibility on it

If not, then I think the best course of action is to log every possible interaction with the db

And I think we probably also should be logging function calls

Like if something doesn't return `Ok(())` then that should get logged

So far this is super helpful, though, as it looks like it already identified an issue

elliottdehn:
@alnoki The problem is that we want to avoid logspam. It's reasonable to log not necessarily every single database interaction but most of them sure. The reason is that it's always possible to go in and add logging once you detect something, versus getting spammed with stuff constantly.

Logging every DB interaction in a loop for example is just going to result in spam as we see 20 times that the aggregator is doing such and such. Let's just log at the top of macroscopic operations and go from there.

alnoki:
@elliottdehn Good point on not overloading the logs

Could we then instead reduce all the logs you have here and consolidate it even more?

E.g. right now it logs every select statement in the user history loop

But instead we could just log at the end of the history loop, either:

* "User history processed and saved", or
* "User history process and save failed: {reason}"

elliottdehn:
We already have this:
> "User history process and save failed: {reason}"
In this form:
```
    #[error("Data could not be processed, reason: {0}")]
    ProcessingError(anyhow::Error),
```
It propagates upward with `?` uses, and eventually we see it.

We could print out successes though, and remove the database interaction logging that I've done. Although we wouldn't have been able to diagnose that bug with such logging alone. I'll go do so to stay in line with processor-style logging.

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -65,6 +69,7 @@ impl Data for UserHistory {
         transaction.execute("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;")
             .await
             .map_err(|e| DataAggregationError::ProcessingError(anyhow!(e)))?;
+
```
alnoki, line 72: How about taking these out?

alnoki: (The blank lines)

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -46,7 +46,10 @@ impl Data for UserHistory {
     }
 
     async fn process_and_save_historical_data(&mut self) -> DataAggregationResult {
-        self.process_and_save_internal().await
+        tracing::info!("[Aggregator] Start processing UserHistory.");
+        let res = self.process_and_save_internal().await;
+        tracing::info!("[Aggregator] Finish processing UserHistory.");
```
alnoki, line 51: I think these two logging statements are redundant since there are already start/end log commands in `process_and_save_interal()`

How about taking these out?

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -46,7 +46,10 @@ impl Data for UserHistory {
...
+        return res;
```
alnoki, line 52: no need to declare `res` only to immediately return it, how about just leaving this function as is?

alnoki (commented):
I left some specific comments, but after looking through the code more I think there is a better way to abstract this

Will discuss offline

src/rust/aggregator/src/data/user_history.rs
elliottdehn: reverted this file

src/rust/aggregator/src/data/user_history.rs
elliottdehn: reverted the file

src/rust/aggregator/src/data/user_history.rs
elliottdehn: reverted the file

alnoki (approved):
This design looks properly abstracted to me

Tagging @CRBl69 so he can comment on effects of logging on async behavior, if any

PR #541, opened 2023-10-11
[ECO-683] Remove daily markets aggregation

alnoki:
* Take out market registration event aggregation, which was a proof of concept, is not necessary at present, and was causing panics


elliottdehn:
Can we also remove the related code for the aggregation? You've removed it's usage here but haven't removed the code itself, which we should do.

alnoki:
@elliottdehn A task to fix the aggregation logic is tracked internally ECO-684

elliottdehn (approved):

PR #549, opened 2023-10-18
[ECO-726] Add CLI args

CRBl69:
- renamed `Data` to `Pipeline`
- can now pass db url as cli arg
- can now pass an exclusion list or an inclusion list to run processors

The purpose of this PR is so that people can run the aggregator to get data without having to process leaderboards.

I also renamed `Data` to `Pipeline` because it's more accurate, and it was hard to name things with `Data`.

### DIFF
--- a/src/rust/aggregator/src/data/user_history.rs
+++ b/src/rust/aggregator/src/data/user_history.rs
@@ -3,7 +3,7 @@ use bigdecimal::{num_bigint::ToBigInt, BigDecimal, Zero};
 use chrono::{DateTime, Duration, Utc};
 use sqlx::{Executor, PgConnection, PgPool, Postgres, Transaction};
 
-use super::{Processor, ProcessorAggregationResult, ProcessorError};
+use super::{Pipeline, PipelineAggregationResult, PipelineError};
 
 /// Number of bits to shift when encoding transaction version.
 const SHIFT_TXN_VERSION: u8 = 64;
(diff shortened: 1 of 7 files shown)


alnoki (changes requested):
As discussed offline

1. How about switching `Processor` to `Pipeline` since we are already using the term processor for our indexer?
2. How about a description that gives concrete examples of running the aggregator in different ways?

alnoki (changes requested):
As discussed offline how about?

1. ~Use #559 as base branch~
2. Resolving the merge conflict

src/rust/aggregator/src/main.rs
```diff
@@ -1,14 +1,58 @@
...
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+struct Args {
+    /// If set, no pipeline will be included by default.
+    #[arg(short, long)]
+    no_default: bool,
+
+    /// Exclusion list. Pipelines specified here will not be executed. Ignored if --no-default passed.
+    #[arg(short, long)]
+    exclude: Option<Vec<Pipelines>>,
```
alnoki, line 23: This is unused

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -49,27 +49,27 @@ impl Data for UserHistory {
...
+    async fn process_and_save_historical_data(&mut self) -> PipelineAggregationResult {
         self.process_and_save_internal().await
     }
 
     fn poll_interval(&self) -> Option<std::time::Duration> {
         Some(std::time::Duration::from_secs(5))
     }
 
-    /// All database interactions are handled in a single atomic transaction. Processor insertions
+    /// All database interactions are handled in a single atomic transaction. Pipeline insertions
```
alnoki: I think this should have been left as the word processor

Because it is indeed referring to the submodule

alnoki (approved):

PR #554, opened 2023-10-21
[ECO-781] optimize queries using last txn_version

CRBl69:
Use the last transaction number instead of a table storing events' primary keys. This is possible because all transactions are inserted at once (not half of a transaction, then the rest) and in order (transaction X is always inserted after transaction X-1).


src/rust/dbv2/migrations/2023-10-06-164142_leaderboards/up.sql
```diff
@@ -99,9 +99,8 @@ SELECT
 CREATE TABLE
   aggregator.competition_indexed_events (
     "txn_version" NUMERIC NOT NULL,
-    "event_idx" NUMERIC NOT NULL,
-    "competition_id" INT NOT NULL REFERENCES aggregator.competition_metadata ("id"),
-    PRIMARY KEY ("txn_version", "event_idx", "competition_id")
+    "competition_id" INT NOT NULL UNIQUE REFERENCES aggregator.competition_metadata ("id"),
```
alnoki, line 102: Is this specifying that `competition_id` must be unique for each row in the table?

alnoki: Perhaps this should be `event_idx`?

alnoki: Why are we getting rid of `event_idx`?

alnoki: Is this table now essentially tracking the transaction versions that have been indexed, under the assumption that all events for each txn are necessarily indexed?

alnoki: Okay I think I get it

This table is saying "the most recently indexed transaction version for a given competition ID"

I found the table name misleading at first but this makes sense now after seeing the other SQL

alnoki (approved):

PR #555, opened 2023-10-21
[ECO-786] improve logging

CRBl69:
Remove duplicated text in logs and add timed logs.


alnoki (changes requested):
The logic looks good to me, but there are some housekeeping issues that should be a quick fix:

1. Remove the SQLx changes, which are breaking cargo check
2. The proposed submodule change is for the wrong commit, so update it to the head of the `econia` branch (63c43d9) which will also take care of ECO-762

alnoki (approved):

PR #558, opened 2023-10-23
[ECO-790] Optimize user history queries

CRBl69:


src/rust/aggregator/src/data/user_history.rs
```diff
@@ -70,227 +70,166 @@ impl Data for UserHistory {
...
-                WHERE NOT EXISTS (
-                    SELECT * FROM aggregator.aggregated_events
-                    WHERE cancel_order_events.txn_version = aggregated_events.txn_version
-                    AND cancel_order_events.event_idx = aggregated_events.event_idx
-                )
+                INSERT INTO aggregator.user_history_limit
+                SELECT
+                    market_id,
+                    order_id,
+                    user,
```
alnoki, line 115: `user` should be quoted since this is a keyword

alnoki: Note that I caught this by moving the SQL into a separate query file and autoformatting on save with `pgformatter`, which made this all caps

src/rust/aggregator/src/data/user_history.rs
```diff
@@ -70,227 +70,166 @@ impl Data for UserHistory {
...
+            r#"
+                INSERT INTO aggregator.user_history
+                SELECT
+                    psoe.market_id,
+                    psoe.order_id,
+                    psoe."time",
+                    NULL,
+                    psoe.integrator,
+                    0,
+                    psoe.max_base * mre.lot_size,
```
alnoki, line 221: This should be division, not multiplication

src/rust/aggregator/src/data/user_history.rs
alnoki: Fixed in #563

src/rust/aggregator/src/data/user_history.rs
alnoki: Fixed in #563

alnoki (changes requested):
Please merge #563

alnoki (approved):

PR #563, opened 2023-10-24
[ECO-800] Abstract SQL, fix keyword and overflow

alnoki:
Clean up #558:
* Use common table expressions for positional parameter readability
* Double quote "user" SQL key word
* Split raw SQL strings into separate files
* Format SQL files on autosave with `pgformatter` extension for VS Code
* Fix error in swap size calculation where multiplication was used instead of division, causing an overflow


CRBl69 (changes requested):
Everything looks good, except the file naming for files `parse_order_events_*.sql`. I would rename them `insert_user_history_*_special.sql` as it does the same as `insert_user_history_*.sql` but for the special fields that are not common to all order types.

PR #565, opened 2023-10-24
[ECO-770] Guard against negative volume

CRBl69:
This change acts as a failsafe in case volume would become negative. This should never happen, but we never know, so it's better to do this than let the aggregator crash.


alnoki:
Closing due to competition over and not wanting to retroactively modify

PR #566, opened 2023-10-24
[ECO-788] Remove all cargo warnings

CRBl69:
This PR removes all cargo warnings. In order to do so, multiple changes were made:

- Move the `Data` type into the `aggregator` library, and use it in the `aggregator` binary.
- Use `#[allow(dead_code)]` where code is not used but might be useful in the future.
- Remove some unused code.
- Move data pipelines into a folder named `pipelines`, since `data` is now part of the `aggregator` library.

There is also a second commit containing a `cargo fmt`.


alnoki (changes requested):
As discussed offline how about?

1. ~Use #559 as base branch~
2. Resolving the merge conflict

alnoki (approved):

PR #582, opened 2023-11-02
[ECO-844] Add price info queries

CRBl69:

PR #588, opened 2023-11-03
[ECO-852] Make price info query suggestions

alnoki:
* Address comments from https://github.com/econia-labs/econia/pull/582
  * Fix `last_fill` and `first_fill` logic
  * Remove lot size multipliers
  * Change percentage to be a relative amount
* Run through `pgformatter` via format on save


CRBl69 (approved):
Looks good :+1:

Let's make sure we run formats in separate commits, so it's easier to review

PR #594, opened 2023-11-05
[ECO-858] create `markets` and `recognized_markets` endpoints

CRBl69:
This PR adds the following endpoints:

- `/markets`
- `/recognized_markets`

You might ask why `/recognized_markets` exists, as you could query them only by
adding `?is_recognized=eq.true` to the first endpoint to get recognized
markets. The reason it that the second endpoint is more performant, as it is
not joining two tables, as would the query mentioned before do.

---

## Example

**Output of querying `/markets`**:

```json
[
  {
    "market_id": 2,
    "time": "2023-09-04T15:20:12.12123+00:00",
    "base_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "base_module_name": "example_apt",
    "base_struct_name": "ExampleAPT",
    "base_name_generic": null,
    "quote_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "quote_module_name": "example_usdc",
    "quote_struct_name": "ExampleUSDC",
    "lot_size": 100000,
    "tick_size": 1,
    "min_size": 1,
    "underwriter_id": 0,
    "is_recognized": false
  },
  {
    "market_id": 3,
    "time": "2023-09-04T17:09:37.423851+00:00",
    "base_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "base_module_name": "example_apt",
    "base_struct_name": "ExampleAPT",
    "base_name_generic": null,
    "quote_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "quote_module_name": "example_usdc",
    "quote_struct_name": "ExampleUSDC",
    "lot_size": 100000,
    "tick_size": 1,
    "min_size": 500,
    "underwriter_id": 0,
    "is_recognized": true
  }
]
```

**Output of querying `/recognized_markets`**:

```json
[
  {
    "time": "2023-10-03T18:27:47.224021+00:00",
    "base_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "base_module_name": "example_apt",
    "base_struct_name": "ExampleAPT",
    "base_name_generic": null,
    "quote_account_address": "0xc0de11113b427d35ece1d8991865a941c0578b0f349acabbe9753863c24109ff",
    "quote_module_name": "example_usdc",
    "quote_struct_name": "ExampleUSDC",
    "market_id": 3,
    "lot_size": 100000,
    "tick_size": 1,
    "min_size": 500,
    "underwriter_id": 0
  }
]
```

### DIFF
--- a/src/rust/dbv2/migrations/2023-11-05-124149_recognized_market_endpoint/up.sql
+++ b/src/rust/dbv2/migrations/2023-11-05-124149_recognized_market_endpoint/up.sql
@@ -1,0 +1,58 @@
+-- Your SQL goes here
+CREATE VIEW api.recognized_markets AS
+    SELECT
+    DISTINCT ON (
+        base_account_address,
+        base_module_name,
+        base_struct_name,
+        base_name_generic,
+        quote_account_address,
+        quote_module_name,
+        quote_struct_name
+    )
+        *
+    FROM
+        recognized_market_events
+    ORDER BY
+        base_account_address,
+        base_module_name,
+        base_struct_name,
+        base_name_generic,
+        quote_account_address,
+        quote_module_name,
+        quote_struct_name,
+        txn_version DESC,
+        event_idx DESC;
+
+
+GRANT SELECT ON api.recognized_markets TO web_anon;
+
+
+CREATE VIEW api.markets AS
+    SELECT
+        m.*,
+        CASE
+            WHEN r.market_id = m.market_id THEN true
+            ELSE false
+        END AS is_recognized
+    FROM
+        market_registration_events AS m
+    LEFT JOIN
+        api.recognized_markets AS r
+    ON
+        COALESCE(r.base_account_address, '') = COALESCE(m.base_account_address, '')
+    AND
+        COALESCE(r.base_module_name, '') = COALESCE(m.base_module_name, '')
+    AND
+        COALESCE(r.base_struct_name, '') = COALESCE(m.base_struct_name, '')
+    AND
+        COALESCE(r.base_name_generic, '') = COALESCE(m.base_name_generic, '')
+    AND
+        r.quote_account_address = m.quote_account_address
+    AND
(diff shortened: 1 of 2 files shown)


src/rust/dbv2/migrations/2023-11-05-124149_recognized_market_endpoint/up.sql
```diff
@@ -0,0 +1,58 @@
...
+    DISTINCT ON (
+        base_account_address,
+        base_module_name,
+        base_struct_name,
+        base_name_generic,
+        quote_account_address,
+        quote_module_name,
+        quote_struct_name
+    )
+        *
```
alnoki, line 13: I think this is missing the additional condition that market_id is not null

Because a `MarketRegistrationEvent` that *de-lists* will have the base/quote fields non-null, but market ID, lot_size, tick_size, min_size, and underwriter_id will be null

The easiest way to verify that the event is a listing event instead of a delisting event is to enforce that market ID is non-null

alnoki (changes requested):
I think the description might be invalid: should one of the example queries actually be for `recognized_market_events()`?

src/rust/dbv2/migrations/2023-11-05-124149_recognized_market_endpoint/up.sql
CRBl69: I see what you mean and it's partially true.

If I add `market_id IS NOT NULL`, and then I query this, I'll ignore the latest delisting event, and I'll return a market that has been delisted.

But leaving it as is will return rows which contain `market_id = NULL`, which might not be what we want.

I'll update

alnoki (approved):

PR #614, opened 2023-11-10
[ECO-829] Include pagination in view function helpers

elliottdehn:
We've added pagination to "get open orders" and "get price levels" so we need to include these functions in the SDK view helpers. I've also included a helper to get a fixed number of orders/price levels that uses the pagination feature. This is a step away from what we've been doing which is to only provide wrappers for the functions and nothing else. However, it's tricky to do the pagination right (have to use the transaction version) so I figured we should just include it as nobody gets hurt.

Validated by calling each paginator in the middle of `trade.py`

### DIFF
--- a/src/python/sdk/econia_sdk/view/market.py
+++ b/src/python/sdk/econia_sdk/view/market.py
@@ -346,7 +346,11 @@ def get_open_orders_paginated(
 
 
 def get_open_orders_with_pagination(
-    viewer: EconiaViewer, market_id: int, max_levels: int = 100, page_size: int = 100
+    viewer: EconiaViewer,
+    market_id: int,
+    max_asks: int = 100,
+    max_bids: int = 100,
+    page_size: int = 100,
 ) -> dict:
     info = viewer.aptos_client.info()
     ledger_version = int(info["ledger_version"])
@@ -363,8 +367,8 @@ def get_open_orders_with_pagination(
         returns = get_open_orders_paginated(
             viewer,
             market_id,
-            min(ask_page_size, max(max_levels - len(asks), 0)),
-            min(bid_page_size, max(max_levels - len(bids), 0)),
+            min(ask_page_size, max(max_asks - len(asks), 0)),
+            min(bid_page_size, max(max_bids - len(bids), 0)),
             int(next_ask_idx),
             int(next_bid_idx),
             ledger_version,
@@ -379,9 +383,9 @@ def get_open_orders_with_pagination(
         for bid in value["bids"]:
             bids.append(_convert_open_order_value(bid))
 
-        if next_ask_idx == "0" and next_bid_idx == "0":
-            keep_looping = False
-        elif len(asks) >= max_levels or len(bids) >= max_levels:
+        if (len(asks) >= max_asks or next_ask_idx == "0") and (
+            len(bids) >= max_bids or next_bid_idx == "0"
+        ):
             keep_looping = False
         elif next_ask_idx == "0" or next_bid_idx == "0":
             if next_ask_idx == "0":
@@ -418,7 +422,11 @@ def get_price_levels_paginated(
 
 
 def get_price_levels_with_pagination(
-    viewer: EconiaViewer, market_id: int, max_levels: int = 100, page_size: int = 100
+    viewer: EconiaViewer,
+    market_id: int,
+    max_asks: int = 100,
+    max_bids: int = 100,
+    page_size: int = 100,
 ) -> dict:
     info = viewer.aptos_client.info()
     ledger_version = int(info["ledger_version"])
@@ -427,16 +435,18 @@ def get_price_levels_with_pagination(
     next_bid_idx = "0"
     bid_page_size = ask_page_size = page_size
 
-    asks = []
-    bids = []
+    asks = dict()
+    asks_size = 0
+    bids = dict()
+    bids_size = 0
 
     keep_looping = True
     while keep_looping:
         returns = get_price_levels_paginated(
             viewer,
             market_id,
-            min(ask_page_size, max(max_levels - len(asks), 0)),
-            min(bid_page_size, max(max_levels - len(bids), 0)),
+            ask_page_size,
+            bid_page_size,
             int(next_ask_idx),
             int(next_bid_idx),
             ledger_version,
@@ -447,13 +457,25 @@ def get_price_levels_with_pagination(
         next_bid_idx = returns[2]
 
         for ask in value["asks"]:
-            asks.append({"price": int(ask["price"]), "size": int(ask["size"])})
+            price = int(ask["price"])
+            size = int(ask["size"])
+            if price in asks:
+                asks[price] += size
+            else:
+                asks[price] = size
         for bid in value["bids"]:
-            bids.append({"price": int(bid["price"]), "size": int(bid["size"])})
+            price = int(bid["price"])
+            size = int(bid["size"])
+            if price in bids:
+                bids[price] += size
+            else:
+                bids[price] = size
+        asks_size += ask_page_size
+        bids_size += bid_page_size
 
-        if next_ask_idx == "0" and next_bid_idx == "0":
-            keep_looping = False
-        elif len(asks) >= max_levels or len(bids) >= max_levels:
+        if (asks_size >= max_asks or next_ask_idx == "0") and (
+            bids_size >= max_bids or next_bid_idx == "0"
+        ):
             keep_looping = False
         elif next_ask_idx == "0" or next_bid_idx == "0":
             if next_ask_idx == "0":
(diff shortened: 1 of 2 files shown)


src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,150 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+            str(next_ask_idx),
+            str(next_bid_idx),
+        ],
+        ledger_version=ledger_version,
+    )
+    return returns
+
+
+def get_open_orders_with_pagination(
+    viewer: EconiaViewer, market_id: int, max_levels: int = 100, page_size: int = 100
```
alnoki, line 349: Rather than `max_levels`, how about `n_max_bids` and `n_max_asks`?

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,150 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+            str(next_ask_idx),
+            str(next_bid_idx),
+        ],
+        ledger_version=ledger_version,
+    )
+    return returns
+
+
+def get_price_levels_with_pagination(
+    viewer: EconiaViewer, market_id: int, max_levels: int = 100, page_size: int = 100
```
alnoki, line 421: How about max levels per side?

There might be situations where people just want asks or bids

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,150 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+        )
+
+        value = returns[0]
+        next_ask_idx = returns[1]
+        next_bid_idx = returns[2]
+
+        for ask in value["asks"]:
+            asks.append({"price": int(ask["price"]), "size": int(ask["size"])})
+        for bid in value["bids"]:
+            bids.append({"price": int(bid["price"]), "size": int(bid["size"])})
```
alnoki, line 452: This doesn't work for returns that break price levels across pages

E.g. if you are paginating you can break in the middle of a price level and the result should be properly accounted for

For example say 3 asks all of size 5 at price 25, but the pagination boundary is two orders in

One pagination call will return 10 total size, the next 15 total size

And in this case there should be logic that consolidates to 25 total size, rather than having two entries for the same price

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,150 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+
+        if next_ask_idx == "0" and next_bid_idx == "0":
+            keep_looping = False
+        elif len(asks) >= max_levels or len(bids) >= max_levels:
+            keep_looping = False
+        elif next_ask_idx == "0" or next_bid_idx == "0":
+            if next_ask_idx == "0":
+                ask_page_size = 0
+            else:
+                bid_page_size = 0
```
alnoki, line 390: This will return early if too many of one side are indexed but not enough of the other, which disregard's the user's specific request

Instead, how about deleting the second condition (first `elif`), then moving the first condition (`if`) to come second?

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,150 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
```
alnoki, line 462: How about the same control flow logic I suggested above?

alnoki (changes requested):
How about an SDK version bump?

src/python/sdk/econia_sdk/view/market.py
elliottdehn: done

src/python/sdk/econia_sdk/view/market.py
elliottdehn: done

src/python/sdk/econia_sdk/view/market.py
elliottdehn: good eye, done

src/python/sdk/econia_sdk/view/market.py
elliottdehn: If we remove the second condition, then we won't stop until both sides are _exhausted_ which is not what we want to do. Even if we did, the first condition must remain the first because the `or` in the third condition would consume the `and` case.

Right now this logic is correct (I think) because both sides have equal max sizes. If they're different, as we're doing now, there needs to be some more logic. Will work on it.

elliottdehn: Handled in most recent revision

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,178 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+                asks[price] = size
+        for bid in value["bids"]:
+            price = int(bid["price"])
+            size = int(bid["size"])
+            if price in bids:
+                bids[price] += size
+            else:
+                bids[price] = size
+        asks_size += ask_page_size
+        bids_size += bid_page_size
```
alnoki, line 474: This is not necessarily true, because even if you ask for a page of size 100 you might only get 50 orders back

So how about incrementing `asks_size` / `bids_size` by the length of the return asks/bids?

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,178 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+    bids = dict()
+    bids_size = 0
+
+    keep_looping = True
+    while keep_looping:
+        returns = get_price_levels_paginated(
+            viewer,
+            market_id,
+            ask_page_size,
+            bid_page_size,
```
alnoki, line 449: If I pass `max_asks = 50, max_bids = 150, page_size = 100`, the function will return 100 asks and 200 bids, correct?

If so, that is a violation of user intent and should be fixed

I believe the open orders with pagination function handles this situation properly and you can probably use similar logic here

src/python/sdk/econia_sdk/view/market.py
alnoki: Also, how about naming this `n_indexed_asks` and `n_indexed_bids`?

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,178 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+            else:
+                bid_page_size = 0
+
+    asks_ls = []
+    bids_ls = []
+    for price, size in asks.items():
+        asks_ls.append({"price": price, "size": size})
+    for price, size in bids.items():
+        bids_ls.append({"price": price, "size": size})
+    return {"asks": asks_ls, "bids": bids_ls, "market_id": market_id}
```
alnoki, line 492: Is the dict `items()` method sorted?

If not, which I suspect it isn't, then how about sorting by keys (ascending for asks, descending for bids) at some point in this operation?

Also, I'm guessing this can probably be done in a one-liner using a list/dict comprehesion or a zip operator

src/python/sdk/econia_sdk/view/market.py
```diff
@@ -320,6 +320,178 @@ def get_price_levels_all(view: EconiaViewer, market_id: int) -> dict:
...
+        ):
+            keep_looping = False
+        elif next_ask_idx == "0" or next_bid_idx == "0":
+            if next_ask_idx == "0":
+                ask_page_size = 0
+            else:
+                bid_page_size = 0
+
+    asks_ls = []
+    bids_ls = []
```
alnoki, line 487: How about something more descriptive like `ask_levels` / `bid_levels` ?

alnoki (changes requested):
Almost there, this handles nearly all of the issues from before

Since there are so many orders on testnet, how about showing some sample outputs in the PR description once you've addressed comments?

src/python/sdk/econia_sdk/view/market.py
elliottdehn: In this case we can't, because the length of the price levels that we get back is not necessarily equal to the length of the orders it covers. This is basically a way of knowing for sure whether or not we could possibly be over the limit.

Is there a case where we will break before getting all necessary items? I don't think so.

src/python/sdk/econia_sdk/view/market.py
elliottdehn: It's like this because we don't know how many orders are covered by the returned price levels result. So this will indeed sometimes process more orders than desired but I figured as long as we stay within the page size and return  an equal or greater number of processed orders then it's fine.

We don't have a way to do this exactly I don't think.

src/python/sdk/econia_sdk/view/market.py
elliottdehn: Python dictionaries are sorted by item insertion, and even if an item is updated its place in the ordering remains the same.

src/python/sdk/econia_sdk/view/market.py
elliottdehn: done

src/python/sdk/econia_sdk/view/market.py
elliottdehn: Fixed

src/python/sdk/econia_sdk/view/market.py
alnoki: > the length of the price levels that we get back is not necessarily equal to the length of the orders it covers

Ahh, right, good point

> This is basically a way of knowing for sure whether or not we could possibly be over the limit.

Indeed, good call

> Is there a case where we will break before getting all necessary items? I don't think so.

Only if indexing is done, but if that's the case next idx will be 0 and will trigger the break condition anyways

alnoki (approved):

PR #622, opened 2023-11-16
[ECO-914] update `/markets` endpoint to include volume, min, and max

CRBl69:
Added a new aggregator pipeline that is named `UpdateMaterializedView`. This updates the given materialized view every given duration. This will allow easly creating views that can be updated at a set interval with minimal aggregator code.

This PR also adds a materialized view for market volume, min fill, and max fill in the last 24 hours.

### DIFF
--- a/src/rust/aggregator/src/pipelines/update_materialized_view.rs
+++ b/src/rust/aggregator/src/pipelines/update_materialized_view.rs
@@ -1,0 +1,66 @@
+use anyhow::anyhow;
+use chrono::{DateTime, Duration, Utc};
+use sqlx::PgPool;
+
+use aggregator::{Pipeline, PipelineAggregationResult, PipelineError};
+
+pub struct UpdateMaterializedView {
+    pool: PgPool,
+    view_name: String,
+    update_interval: std::time::Duration,
+    last_indexed_timestamp: Option<DateTime<Utc>>,
+}
+
+#[allow(dead_code)]
+impl UpdateMaterializedView {
+    pub fn new(
+        pool: PgPool,
+        view_name: impl Into<String>,
+        update_interval: std::time::Duration,
+    ) -> Self {
+        Self {
+            pool,
+            view_name: view_name.into(),
+            update_interval,
+            last_indexed_timestamp: None,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Pipeline for UpdateMaterializedView {
+    fn model_name(&self) -> String {
+        String::from("UpdateMaterializedView")
+    }
+
+    fn ready(&self) -> bool {
+        self.last_indexed_timestamp.is_none()
+            || self.last_indexed_timestamp.unwrap()
+                + Duration::from_std(self.update_interval).unwrap()
+                < Utc::now()
+    }
+
+    async fn process_and_save_historical_data(&mut self) -> PipelineAggregationResult {
+        self.process_and_save_internal().await
+    }
+
+    fn poll_interval(&self) -> Option<std::time::Duration> {
+        Some(self.update_interval)
+    }
+
+    async fn process_and_save_internal(&mut self) -> PipelineAggregationResult {
+        sqlx::query(
+            format!(
+                r#"
+                REFRESH MATERIALIZED VIEW {};
+            "#,
+                self.view_name
+            )
+            .as_str(),
+        )
+        .execute(&self.pool)
+        .await
+        .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
+        Ok(())
+    }
+}
--- a/src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
+++ b/src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
@@ -70,8 +70,8 @@ CREATE VIEW api.markets AS
             ELSE false
         END AS is_recognized,
         l.price AS last_fill_price_24hr,
-        (COALESCE(l.price, 1) - COALESCE(f.price, 1)) / COALESCE(f.price, 1) * 100 AS percent_change_24h,
-        l.price - f.price AS nominal_change_24h,
+        (COALESCE(l.price, 1) - COALESCE(f.price, 1)) / COALESCE(f.price, 1) * 100 AS price_change_as_percent_24hr,
+        l.price - f.price AS price_change_24hr,
         v.min_price_24h,
         v.max_price_24h,
         v.base_volume_24h,
--- a/src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
+++ b/src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
@@ -1,0 +1,111 @@
+-- Your SQL goes here
+CREATE MATERIALIZED VIEW aggregator.markets_24h_data AS
+    WITH fills AS(
+        SELECT
+            *
+        FROM
+            fill_events
+        WHERE
+            "time" >= CURRENT_TIMESTAMP - '1 day'::interval
+            AND emit_address = maker_address
+    )
+    SELECT
+        market_id,
+        MIN(fills.price) AS min_price_24h,
+        MAX(fills.price) AS max_price_24h,
+        SUM(fills.size) AS base_volume_24h,
+        SUM(fills.size * fills.price) AS quote_volume_24h
+    FROM
+        fills
+    GROUP BY
+        market_id;
+
+
+DROP VIEW api.markets;
+
+
+CREATE VIEW api.markets AS
+    WITH last_fills AS(
+        SELECT
+        DISTINCT ON (market_id)
+            *
+        FROM
+            fill_events
+        WHERE
+            "time" >= CURRENT_TIMESTAMP - '1 day'::interval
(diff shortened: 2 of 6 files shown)


src/rust/aggregator/src/pipelines/update_materialized_view.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::anyhow;
+use chrono::{DateTime, Duration, Utc};
+use sqlx::PgPool;
+
+use aggregator::{Pipeline, PipelineAggregationResult, PipelineError};
+
+pub struct UpdateMaterializedView {
```
alnoki, line 7: Before we lock this into the codebase, how about we go with `RefreshMaterializedView` , to make it clear that we are doing a 1-to-1 PostgreSQL operation?

src/rust/aggregator/src/pipelines/update_materialized_view.rs
```diff
@@ -0,0 +1,66 @@
...
+    pool: PgPool,
+    view_name: String,
+    update_interval: std::time::Duration,
+    last_indexed_timestamp: Option<DateTime<Utc>>,
+}
+
+#[allow(dead_code)]
```
alnoki, line 14: Why is this here?

src/rust/aggregator/src/pipelines/update_materialized_view.rs
```diff
@@ -0,0 +1,66 @@
...
+            update_interval,
+            last_indexed_timestamp: None,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl Pipeline for UpdateMaterializedView {
+    fn model_name(&self) -> String {
+        String::from("UpdateMaterializedView")
```
alnoki, line 33: `RefreshMaterializedView` ?

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,111 @@
...
+    LEFT JOIN
+        first_fills AS f
+    ON
+        f.market_id = m.market_id
+    LEFT JOIN
+        last_fills AS l
+    ON
+        l.market_id = m.market_id
+    LEFT JOIN
+        aggregator.markets_24h_data AS v
```
alnoki, line 106: If the materialized view hasn't been precomputed, will this operation compute it or error out?

src/rust/aggregator/src/pipelines/update_materialized_view.rs
```diff
@@ -0,0 +1,66 @@
...
+        Some(self.update_interval)
+    }
+
+    async fn process_and_save_internal(&mut self) -> PipelineAggregationResult {
+        sqlx::query(
+            format!(
+                r#"
+                REFRESH MATERIALIZED VIEW {};
+            "#,
+                self.view_name
```
alnoki, line 57: How about making the raw string a one-liner? It doesn't need three

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,111 @@
...
+    AND
+        r.quote_module_name = m.quote_module_name
+    AND
+        r.quote_struct_name = m.quote_struct_name
+    LEFT JOIN
+        first_fills AS f
+    ON
+        f.market_id = m.market_id
+    LEFT JOIN
+        last_fills AS l
```
alnoki, line 102: I don't see a `LIMIT` statement on either first or last fills

Will this be joining more rows than needed?

src/rust/aggregator/src/pipelines/update_materialized_view.rs
```diff
@@ -0,0 +1,66 @@
...
+    async fn process_and_save_historical_data(&mut self) -> PipelineAggregationResult {
+        self.process_and_save_internal().await
+    }
+
+    fn poll_interval(&self) -> Option<std::time::Duration> {
+        Some(self.update_interval)
+    }
+
+    async fn process_and_save_internal(&mut self) -> PipelineAggregationResult {
+        sqlx::query(
```
alnoki, line 52: Do you need to run SQLx prepare for this?

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,111 @@
...
+        m.tick_size,
+        m.min_size,
+        m.underwriter_id,
+        CASE
+            WHEN r.market_id = m.market_id THEN true
+            ELSE false
+        END AS is_recognized,
+        l.price AS last_fill_price_24hr,
+        (COALESCE(l.price, 1) - COALESCE(f.price, 1)) / COALESCE(f.price, 1) * 100 AS percent_change_24h,
+        l.price - f.price AS nominal_change_24h,
```
alnoki, line 74: This is the change in integer prices over 24 hr, whereas prices don't become nominal until coin decimals are accounted for

So how about `AS price_change_24h`?

edit: okay I see that it is nominal to contrast with the percent change field which comes directly before

So instead, how about `AS price_change_24hr`, and the prior one `price_change_as_percent_24hr`

alnoki (changes requested):

src/rust/aggregator/src/pipelines/update_materialized_view.rs
CRBl69: :+1:

src/rust/aggregator/src/pipelines/update_materialized_view.rs
CRBl69: Copy paste error.

src/rust/aggregator/src/pipelines/update_materialized_view.rs
CRBl69: :+1:

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
CRBl69: image

It looks like it gets computed once on creation.

src/rust/aggregator/src/pipelines/update_materialized_view.rs
CRBl69: :+1:

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
CRBl69: That's taken care of by the `DISTINCT ON (market_id)`, which only selects one fill event per market.

src/rust/aggregator/src/pipelines/update_materialized_view.rs
CRBl69: No. This is unchecked. Using `query!` doesn't work here, since it doesn't look like it supports passing table/views names as an argument. And it also doesn't support passing a processed string as an argument, it wants a string literral.

src/rust/aggregator/src/pipelines/update_materialized_view.rs
alnoki: Note this still says `Update...` as of 42ed45f3

I highlighted this one separately before as failing to change it wouldn't produce a compiler error

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,115 @@
...
+    first_fills AS(
+        SELECT
+            *
+        FROM
+            fill_events
+        WHERE
+            "time" >= CURRENT_TIMESTAMP - '1 day'::interval
+        LIMIT 1
+    )
+    SELECT
```
alnoki, line 58: Now that this no longer has distinct on market ID, will the max time and first fill be correctly computed for the market ID when doing the later left join?

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,115 @@
...
+DROP VIEW api.markets;
+
+
+CREATE VIEW api.markets AS
+    WITH max_time AS (
+        SELECT
+            MAX("time") AS mtime
+        FROM
+            fill_events
+    ),
```
alnoki, line 33: If multiple events happen in the same txn, this will pick the earliest one, correct? And thus not necessarily the latest one in terms of emission sequence?

src/rust/dbv2/migrations/2023-11-16-103707_market_volume/up.sql
```diff
@@ -0,0 +1,115 @@
...
+        LIMIT 1
+    ),
+    first_fills AS(
+        SELECT
+            *
+        FROM
+            fill_events
+        WHERE
+            "time" >= CURRENT_TIMESTAMP - '1 day'::interval
+        LIMIT 1
```
alnoki, line 56: How about pinning this to the earliest txn version/event idx?

alnoki (changes requested):

alnoki (approved):

PR #625, opened 2023-11-16
[ECO-904] add coins pipeline

CRBl69:
1. Implement aggregator coin pipeline to store coin name, symbol, and decimals.
1. Add support for and document `APTOS_NETWORK` environment variable required for coin chain queries via Rust SDK.
1. Bump changeloge for v1.4.0, update formatting
1. Rebuild REST API docs
1. Clean up aggregator `clap` implementation via `ValueEnum` derivation


src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+        let coins = sqlx::query!(
+            r#"
+                WITH market_coins AS (
+                    SELECT DISTINCT
+                        base_account_address AS address,
+                        base_module_name AS module,
+                        base_struct_name AS struct
+                    FROM
+                        market_registration_events
+                    WHERE base_account_address IS NOT NULL
```
alnoki, line 66: where `base_name_generic` is null

src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+                    let CoinInfo { name, symbol, decimals } = coin_info;
+                    sqlx::query!(
+                        "INSERT INTO aggregator.coins VALUES ($1, $2, $3, $4, $5, $6)",
+                        name, symbol, decimals, address, module, struct_name
+                    )
+                    .execute(&self.pool)
+                    .await
+                    .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
+                },
+                Err(GetCoinInfoError::RateLimited) => {},
```
alnoki, line 114: How about exiting the loop when rate limited?

alnoki: How about an info log saying we got rate limited?

src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+                Err(GetCoinInfoError::Other(e)) => {return Err(PipelineError::ProcessingError(anyhow!(e)));},
```
alnoki, line 115: How about logging an error here for any error other than a rate limit?

src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+
+#[derive(Debug, Error)]
+enum GetCoinInfoError {
+    #[error("Got rate limited by Aptos")]
+    RateLimited,
+    #[error("Other error: {0}")]
+    Other(anyhow::Error),
+}
+
+async fn get_coin_info(move_struct_tag_str: String) -> Result<CoinInfo, GetCoinInfoError> {
```
alnoki, line 136: To do this, query for the `0x1::coin::CoinInfo<CoinType>` under the coin address, then extract out the `name`, `symbol`, and `decimals` fields

src/rust/aggregator/src/pipelines/coins.rs
alnoki, line 1: This will require an environment variable for the fullnode to query coin info from

How about using Aptos labs testnet fullnode as default?

alnoki: Note this will require updating the dockerfile too

alnoki: And for docker compose, we should make it default to testnet, but override via command line when calling docker compose up

src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+                    )
+                    .execute(&self.pool)
+                    .await
+                    .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
+                },
+                Err(GetCoinInfoError::RateLimited) => {},
+                Err(GetCoinInfoError::Other(e)) => {return Err(PipelineError::ProcessingError(anyhow!(e)));},
+            }
+        }
+        Ok(())
```
alnoki, line 118: We might want to have a table that tracks `aggregator.coins_last_time_rate_limited`

Then we have a constant like `TIME_TO_WAIT_AFTER_GETTING_RATE_LIMITED`

If it hasn't been this long since the last query, do not try querying at all

Once this is implemented, we can run the pipeline on a 1s polling interval to give real-time market registration updates

src/rust/aggregator/src/pipelines/coins.rs
```diff
@@ -0,0 +1,138 @@
...
+    async fn process_and_save_historical_data(&mut self) -> PipelineAggregationResult {
+        sqlx::query!(
+            r#"
+                INSERT INTO aggregator.markets_registered_per_day (markets, date)
+                SELECT COUNT(*), time::date
+                FROM market_registration_events
+                WHERE time::date NOT IN (
+                    SELECT date FROM aggregator.markets_registered_per_day
+                )
+                GROUP BY time::date
```
alnoki, line 43: How about separate SQLx files?

src/rust/aggregator/src/pipelines/coins.rs
alnoki: For this to be easily deployed on cloud compute it's looking like it will need to have an environment variable option in addition to the command line argument

And I'm starting to think that we shouldn't use defaults, because people will get familiar with it on testnet and then have no idea why the coins aren't loading on mainnet

alnoki (approved):

PR #650, opened 2023-12-01
[ECO-981] add fees to user_history

CRBl69:
Add a fee field to the user history and orders enpoint.

### DIFF
--- a/src/rust/aggregator/src/pipelines/user_history.rs
+++ b/src/rust/aggregator/src/pipelines/user_history.rs
@@ -156,6 +156,7 @@ impl Pipeline for UserHistory {
                             &fill.market_id,
                             &fill.time,
                             &fill.price,
+                            &fill.taker_quote_fees_paid,
                         )
                         .await?;
                     }
@@ -201,9 +202,10 @@ async fn aggregate_fill_for_maker_and_taker<'a>(
     market_id: &BigDecimal,
     time: &DateTime<Utc>,
     price: &BigDecimal,
+    fees: &BigDecimal,
 ) -> PipelineAggregationResult {
-    aggregate_fill(tx, size, maker_order_id, market_id, time, price).await?;
-    aggregate_fill(tx, size, taker_order_id, market_id, time, price).await?;
+    aggregate_fill(tx, size, maker_order_id, market_id, time, price, fees).await?;
+    aggregate_fill(tx, size, taker_order_id, market_id, time, price, fees).await?;
     Ok(())
 }
 
@@ -214,6 +216,7 @@ async fn aggregate_fill<'a>(
     market_id: &BigDecimal,
     time: &DateTime<Utc>,
     price: &BigDecimal,
+    fees: &BigDecimal,
 ) -> PipelineAggregationResult {
     // Only limit orders can remain open after a transaction during which they are filled against,
     // so flag market orders and swaps as closed by default: if they end up being cancelled instead
@@ -226,6 +229,7 @@ async fn aggregate_fill<'a>(
         market_id,
         time,
         price,
+        fees,
     )
     .execute(tx as &mut PgConnection)
     .await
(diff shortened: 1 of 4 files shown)


src/rust/aggregator/src/pipelines/user_history.rs
```diff
@@ -201,9 +202,10 @@ async fn aggregate_fill_for_maker_and_taker<'a>(
     market_id: &BigDecimal,
     time: &DateTime<Utc>,
     price: &BigDecimal,
+    fees: &BigDecimal,
 ) -> PipelineAggregationResult {
-    aggregate_fill(tx, size, maker_order_id, market_id, time, price).await?;
-    aggregate_fill(tx, size, taker_order_id, market_id, time, price).await?;
+    aggregate_fill(tx, size, maker_order_id, market_id, time, price, fees).await?;
```
alnoki, line 207: How about passing `0` rather than `fees` for maker side, since makers dont pay fees?

Otherwise, the order history for the maker order will report fees paid as nonzero

src/rust/dbv2/migrations/2023-12-01-173351_add_total_fees/up.sql
```diff
@@ -0,0 +1,36 @@
...
+ADD COLUMN total_fees_paid_in_quote_subunits NUMERIC(20,0);
+
+
+WITH total_fees AS (
+    SELECT
+        SUM(taker_quote_fees_paid) AS total_fees_paid_in_quote_subunits,
+        taker_order_id,
+        market_id
+    FROM fill_events AS f
+    WHERE f.taker_address = f.emit_address
```
alnoki, line 15: How about switching to `f.maker_address`, since a fill event isn't always emitted to taker side (e.g. non-signing swap)?

src/rust/dbv2/migrations/2023-12-01-173351_add_total_fees/up.sql
```diff
@@ -0,0 +1,36 @@
...
+    GROUP BY market_id, taker_order_id
+)
+UPDATE aggregator.user_history AS u
+SET total_fees_paid_in_quote_subunits = f.total_fees_paid_in_quote_subunits
+FROM total_fees AS f
+WHERE f.taker_order_id = u.order_id
+AND f.market_id = u.market_id;
```
alnoki, line 22: After looking through the rest of the logic in this PR, it looks to me like this block was an attempt to backfill fee data, rather than restarting the DSS, during development. But given the other aggregator logic you're proposing, I think you should be able to cut this out

How about updating this migration to the simpler version below:

```sql
DROP VIEW api.orders;


ALTER TABLE aggregator.user_history
ADD COLUMN total_fees_paid_in_quote_subunits NUMERIC(20, 0) NOT NULL;


CREATE VIEW
  api.orders AS
SELECT
  *
FROM
  aggregator.user_history;


GRANT
SELECT
  ON api.orders TO web_anon;
```

alnoki (changes requested):

src/rust/aggregator/src/pipelines/user_history.rs
CRBl69: Right, thanks, updating

src/rust/dbv2/migrations/2023-12-01-173351_add_total_fees/up.sql
CRBl69: Yes, it is supposed to backfill.

I'm not sure about taking this out. We should strive to make the DSS have a better developer experience, and this is part of this.

Let's talk more about this on the daily.

alnoki: Noting here that backfill logic enables hot DSS upgrades, and that we need to also account for last indexed txn version number for the relevant pipeline when constructing hot upgrade migrations

alnoki (approved):

PR #658, opened 2023-12-15
[ECO-977] Add tables needed for liquidity analytics

CRBl69:
This PR adds two aggregator pipelines:

1. Order history, which is a regular historical snapshot of order book state over time
2. Rolling 24 hour volume

Both pipelines generate substantial data, such that the corresponding pipelines are optional. To activate these pipelines, they must be specified either via CLI arguments or environment variables.

To check the CLI arguments required, use `--help` from the aggregator.

Then, to include as an environment variable, use the `+` delimiter

`AGGREGATOR_INCLUDE=order-history+rolling-volume`


alnoki:
Per https://github.com/econia-labs/econia/pull/659#issuecomment-1858420925

- [ ] Merge this to a `dss-v1.5.0` branch with to use the environment variable file config option (showing "additional-pipelines")

src/rust/aggregator/src/pipelines/order_history.rs
```diff
@@ -0,0 +1,324 @@
...
+                *index += 1;
+                let order_key_maker = (fill.market_id.clone(), fill.maker_order_id.clone());
+                if let Some(order) = orders.get_mut(&order_key_maker) {
+                    order.size -= fill.size.clone();
+                    if order.size.is_zero() {
+                        orders.remove(&order_key_maker);
+                    }
+                }
+                let order_key_taker = (fill.market_id.clone(), fill.taker_order_id.clone());
+                if let Some(order) = orders.get_mut(&order_key_taker) {
```
alnoki, line 173: This only gets called in the case of a limit order that takes, correct?

src/rust/aggregator/sqlx_queries/order_history/insert_order.sql
```diff
@@ -0,0 +1,15 @@
+WITH parameters AS (
+    SELECT
+        $1::numeric(20,0) AS market_id,
+        $2::numeric(39,0) AS order_id,
+        $3::text AS user,
+        $4::numeric(20,0) AS size,
```
alnoki, line 6: How about `"size"`?

src/rust/aggregator/sqlx_queries/order_history/insert_order.sql
```diff
@@ -0,0 +1,15 @@
...
+        $5::order_direction AS direction,
+        $6::numeric(20,0) AS price,
+        $7::timestamptz AS time
```
alnoki, line 9: How about `"time"`?

src/rust/aggregator/src/pipelines/order_history.rs
```diff
@@ -0,0 +1,324 @@
...
+            sqlx::query_file!(
+                "sqlx_queries/order_history/update_last_indexed_timestamp.sql",
+                last,
+            )
+            .execute(&mut transaction as &mut PgConnection)
+            .await
+            .map_err(to_pipeline_error)?;
+        }
+        commit_transaction(transaction).await?;
+        self.state = Some(state);
```
alnoki, line 321: How does this handle hot restarts/upgrades?

src/rust/aggregator/src/pipelines/order_history.rs
```diff
@@ -0,0 +1,324 @@
...
+            }
+        }
+        while fills_index < fills.len() {
+            let fill = fills.get(fills_index).unwrap();
+            handle_fill(fill, &mut orders, &mut fills_index);
+        }
+        while changes_index < changes.len() {
+            let change = changes.get(changes_index).unwrap();
+            handle_change(change, &mut orders, &mut changes_index);
+        }
```
alnoki, line 211: I think this loop could be more effectively abstracted to only have one call each to `handle_fill` and `handle_change`

How about making it more like the `user_history` call that starts

```rust
for _ in 0..(fill_events.len() + change_events.len())
```

src/rust/aggregator/src/pipelines/order_history.rs
```diff
@@ -0,0 +1,324 @@
...
+                    state.orders,
+                    &state.last_txn_indexed,
+                    timestamp,
+                    &mut transaction,
+                )
+                .await?
+            };
+            for order in state.orders.values() {
+                sqlx::query_file!(
+                    "sqlx_queries/order_history/insert_order.sql",
```
alnoki, line 297: Could this end up inserting a taker order?

src/rust/aggregator/src/main.rs
```diff
@@ -195,6 +198,8 @@ async fn main() -> Result<()> {
                     Duration::from_secs(5 * 60),
                 ))))
             }
+            Pipelines::OrderHistory => data.push(Arc::new(Mutex::new(OrderHistory::new(pool.clone())))),
```
alnoki, line 201: How about adding in env var logic that handles both `clap` and env var inputs, similar to how `APTOS_NETWORK` is handled above?

src/rust/aggregator/src/pipelines/leaderboards.rs
```diff
@@ -76,10 +76,7 @@ impl Pipeline for Leaderboards {
         for comp in competitions {
             aggregate_data_for_competition(&mut transaction, comp).await?;
         }
-        transaction
-            .commit()
-            .await
-            .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
+        commit_transaction(transaction).await?;
```
alnoki, line 79: If you're going to update this, how about also abstracting in the new util function to create repeatable read txn?

src/rust/aggregator/src/pipelines/candlesticks.rs
```diff
@@ -76,10 +76,7 @@ impl Pipeline for Candlesticks {
         .await
         .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
 
-        transaction
-            .commit()
-            .await
-            .map_err(|e| PipelineError::ProcessingError(anyhow!(e)))?;
+        commit_transaction(transaction).await?;
```
alnoki, line 79: If you're going to update this, how about also abstracting in the new util function to create repeatable read txn?

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
```diff
@@ -0,0 +1,89 @@
...
+
+CREATE TABLE aggregator.order_history_last_indexed_timestamp (
+    "time" timestamptz PRIMARY KEY
+);
+
+CREATE TABLE aggregator.daily_rolling_volume_history_last_indexed_timestamp (
+    "time" timestamptz PRIMARY KEY
+);
+
+CREATE FUNCTION get_denom_for_market(numeric) RETURNS NUMERIC IMMUTABLE AS $$
```
alnoki, line 71: I mostly understand what's going on here, but it doesn't look like this gets used anywhere

Also, this applies only to quote, not base, so if anything it would be `get_quote_volume_divisor_for_market`, right?

src/rust/aggregator/src/pipelines/rolling_volume.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    fn poll_interval(&self) -> Option<std::time::Duration> {
+        Some(TIMEOUT)
+    }
+
+    async fn process_and_save_internal(&mut self) -> PipelineAggregationResult {
+        let mut transaction = create_repeatable_read_transaction(&self.pool).await?;
+        sqlx::query_file!(
+            "sqlx_queries/rolling_volume/insert_daily_rolling_volume.sql",
```
alnoki, line 45: Can we replace this pipeline with a view function that just gets the values from 24 hr candlesticks?

src/rust/aggregator/src/util.rs
```diff
@@ -0,0 +1,27 @@
...
+) -> Result<Transaction<'a, Postgres>, PipelineError> {
+    let mut transaction = pool.begin().await.map_err(to_pipeline_error)?;
+    transaction
+        .execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;")
+        .await
+        .map_err(to_pipeline_error)?;
+    Ok(transaction)
+}
+
+pub async fn commit_transaction<'a>(tx: Transaction<'a, Postgres>) -> PipelineAggregationResult {
```
alnoki, line 22: Note that repeatable read can still have serialization errors: Repeatable read can still through txn errors: https://www.postgresql.org/docs/current/transaction-iso.html#XACT-REPEATABLE-READ

If that's the case, how about returning an error to the caller?

Then, depending on the pipeline, either gracefully igoring or otherwise

alnoki: Alternatively it could retry, like in the processor

alnoki (changes requested):

src/rust/aggregator/src/pipelines/user_history.rs
```diff
@@ -187,20 +172,7 @@ impl Pipeline for UserHistory {
...
-                    if dbe.message() == "could not serialize access due to read/write dependencies among transactions" {
-                        warn!("transaction serialization error, gracefully ignoring, will retry next loop");
-                        Ok(())
-                    } else {
-                        Err(PipelineError::ProcessingError(anyhow!(e)))
-                    }
-                }
-                _ => {Err(PipelineError::ProcessingError(anyhow!(e)))}
-            })?;
+        commit_transaction(transaction).await?;
```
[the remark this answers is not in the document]
CRBl69: TBH, I'm not sure how to handle this. After investigating the last one, I figured out the error was coming from a `sqlx::query` call and not a `commit` call, so I don't think adding anything here will prevent errors.

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
```diff
@@ -0,0 +1,89 @@
...
+    SELECT tick_size * POW(10::numeric,decimals::numeric)
+    FROM market_registration_events AS m
+    INNER JOIN aggregator.coins AS c
+    ON m.quote_account_address = c."address"
+    AND m.quote_module_name = c.module
+    AND m.quote_struct_name = c.struct
+    WHERE market_id = $1;
+$$ LANGUAGE sql;
+
+CREATE ROLE grafana ENCRYPTED PASSWORD 'grafana' LOGIN;
```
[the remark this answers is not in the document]
CRBl69: I don't think so. Best alternative would be:

- do not set a password and do not allow login
- have a script at CI time run the following commands:
  - `ALTER ROLE grafana ENCRYPTED PASSWORD 'generated_pw';`
  - `ALTER ROLE grafana LOGIN;`

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
```diff
@@ -0,0 +1,89 @@
...
+maxes AS (
+    SELECT * FROM place_max
+    UNION
+    SELECT * FROM cancel_max
+    UNION
+    SELECT * FROM change_max
+    UNION
+    SELECT * FROM fill_max
+)
+SELECT MAX("time") AS time FROM maxes;
```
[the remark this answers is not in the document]
CRBl69: After `AS`, they are not really useful, since there's no confusion that can happen, but I can still add them for consistency.

src/rust/aggregator/src/pipelines/rolling_volume.rs
```diff
@@ -0,0 +1,67 @@
...
+        sqlx::query_file!(
+            "sqlx_queries/rolling_volume/delete_last_indexed_timestamp.sql",
+        )
+        .execute(&mut transaction as &mut PgConnection)
+        .await
+        .map_err(to_pipeline_error)?;
+        sqlx::query_file!(
+            "sqlx_queries/rolling_volume/update_last_indexed_timestamp.sql",
+        )
+        .execute(&mut transaction as &mut PgConnection)
```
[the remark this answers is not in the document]
CRBl69: You cannot put two statements in a prepared query. I did and got an error.

src/rust/aggregator/sqlx_queries/order_history/init_last_indexed_timestamp.sql
```diff
@@ -0,0 +1,10 @@
+-- We can assume that everything before this timestamp was indexed, since
+-- there's nothing to index.
+INSERT INTO aggregator.order_history_last_indexed_timestamp ("time")
+SELECT
+    DATE_TRUNC('minute', "time")
+FROM
+    place_limit_order_events
+ORDER BY
+    "time"
+LIMIT 1
```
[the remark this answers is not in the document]
CRBl69: `ASC` is ALWAYS the default.

src/rust/aggregator/sqlx_queries/order_history/init_last_indexed_timestamp.sql
```diff
@@ -0,0 +1,10 @@
+-- We can assume that everything before this timestamp was indexed, since
+-- there's nothing to index.
+INSERT INTO aggregator.order_history_last_indexed_timestamp ("time")
+SELECT
+    DATE_TRUNC('minute', "time")
+FROM
+    place_limit_order_events
```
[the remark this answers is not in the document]
CRBl69: If there's no limit orders, this will return 0 rows, so nothing will be inserted, and no error would be raised. In the [Rust code](https://github.com/econia-labs/econia/pull/658/files/03a510d21ecbaff9dcc00fbf2e23f07779a2ce54#diff-e5ca5ff9be33fe28c5ce766144fe36c6af89372639b8002835f64e9be74e08e5R243-R256), this case is handled as well.

src/rust/aggregator/src/pipelines/order_history.rs
CRBl69: This closure gets called for every deduped fill. The two `if`s are here to ensure we did not already cancel the order in the `for` above.

src/rust/aggregator/src/pipelines/order_history.rs
CRBl69: If the state is empty, it will calculate it from scratch using the events in the database (see [here](https://github.com/econia-labs/econia/pull/658/files/03a510d21ecbaff9dcc00fbf2e23f07779a2ce54#diff-e5ca5ff9be33fe28c5ce766144fe36c6af89372639b8002835f64e9be74e08e5R278-R286)).

This means that on aggregator startup, the first time this runs, it will take a little longer to calculate the state, but for following runs, it will use this precalculated state to generate the following states.

src/rust/aggregator/src/pipelines/order_history.rs
CRBl69: Taker orders are closed in the same transaction where they are open right ? If yes, and there's no other logic error in the code, then no.

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
CRBl69: This is a utility function later used in the Grafana queries.

Will update the name.

src/rust/aggregator/src/pipelines/rolling_volume.rs
CRBl69: We would lose a lot of precision.

The current code calculates the 24h volume for each minute. So that means we know the volume that was traded between 00h00 to 00h00 the next day, 00h01 to 00h01 the next day, et cetera.

Using candlesticks would only give us the volume between 00h00 to 00h00 the next day.

Having it minute to minute is useful to create graphs like this:

image

rather than like this:

image

src/rust/aggregator/src/pipelines/user_history.rs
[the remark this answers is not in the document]
alnoki: Got it, well if we successully get the processor and aggregator to hot restart then if they do crash I suppose Cloud Run should have them pick up right where they left off anyways, correct?

src/rust/aggregator/src/pipelines/order_history.rs
alnoki: The new logic is more abstracted, but it still has two calls each to `handle_fill` and `handle_change` that could be consolidated for easier maintainability

How about using the `let (fill_event_to_aggregate, change_event_to_aggregate)` paradigm as in the original algorithm, ending with the `mach(fill_event_to_aggregator, change_event_to_aggregate)` statement that would, in this case, result in a *single* call to `handle_fill` and `handle_change`?

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
```diff
@@ -0,0 +1,89 @@
+-- Your SQL goes here
+CREATE VIEW aggregator.last_time_indexed AS
```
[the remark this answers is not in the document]
alnoki: So now it is `latest_event_time`, which is fine within the context of the order history pipeline

But for the context of the aggregator schema, `aggregator.latest_event_timestamp` is confusing, because this does *not* take into account the latest event timestamp from size changes, market orders, etc.

So how about `aggregator.order_history_latest_event_timestamp`?

src/docker/aggregator/Dockerfile
```diff
@@ -41,4 +41,4 @@ COPY --from=builder $BIN_PATH $APP_DIR/
 ENV APTOS_NETWORK=$APTOS_NETWORK
 ENV DATABASE_URL=$DATABASE_URL
 
-ENTRYPOINT /app/aggregator
+ENTRYPOINT /app/aggregator --include rolling-volume --include order-history
```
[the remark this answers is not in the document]
alnoki: This should be taken out, so that people running local docker compose don't always have the expensive rolling volume and order history pipelines

src/rust/aggregator/src/main.rs
CRBl69: I did this. You can see above that I slightly reworked how env vars and args work. Both are taken into account, but env vars always takes priority.

src/rust/aggregator/src/pipelines/user_history.rs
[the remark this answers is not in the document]
CRBl69: Yes

src/rust/aggregator/src/pipelines/order_history.rs
CRBl69: I'm sorry but I do not get what you mean. Furthermore, since `handle_*` are functions, I do not see how calling them twice would affect maintainability.

src/rust/aggregator/src/util.rs
CRBl69: As noted above, what happened to us with the weird "maybe it will succeed if retried" error, it is not when commiting that the error occurs, but on individual calls.

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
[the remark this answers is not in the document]
alnoki: How about changing this to `grafana_read_only` and password `grafana_read_only`?

This shouldn't be an issue with our private networking stack on GCP as long as we firewall out public IP addresses

And maybe add a comment?

```
--This compromised password is not problematic for a cloud deployment with private IP networking
```

src/rust/aggregator/src/pipelines/order_history.rs
alnoki: Noting offline discussion to incorporate the `_to_aggregate` option paradigm for consolidation of duplicate function calls

src/rust/aggregator/src/main.rs
alnoki: Per offline discussion, the comma schema is problematic because environment variables in GCP are parsed of form

`APTOS_NETWORK=my_network,DATABASE_URL,PIPELINES=foo,bar,baz`

Hence a different delimiter is necessary

- [ ] Use a different delimiter, `/`, which we know is valid due to its inclusion in paths/urls as env vars
- [ ] Use a special parser/datatype etc. that gets a vector of pipelines from an env var (this applies to `exclude` and `include` options)
- [ ] Include the `no_default` option
- [ ] Make aggregator-specific ones of form `AGGREGATOR_INCLUDE`, `AGGREGATOR_EXCLUDE`, `AGGREGATOR_NO_DEFAULT`

src/rust/aggregator/src/util.rs
alnoki: Noting that the error is from a query, not form the commit operation

Also, this is handled by using a container with auto-restart

src/rust/dbv2/migrations/2023-12-07-135237_metrics/up.sql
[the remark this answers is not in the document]
alnoki: See https://github.com/econia-labs/econia/blob/76f2c373f6bae6c13c515b2f78b05d9937d74be5/src/rust/dbv2/migrations/2023-12-19-000000_roles/up.sql

PR #659, opened 2023-12-15
[ECO-1055] make processor easly restartable

CRBl69:
Make the processor restartable by getting the latest indexed txn from the db on startup.

Also copy instead of mount the config file.

### DIFF
--- a/src/docker/processor/Dockerfile
+++ b/src/docker/processor/Dockerfile
@@ -61,5 +61,8 @@ ENV RUST_LOG_FORMAT=json
 # Health check port.
 EXPOSE 8084
 
+# This helps the processor stop faster
+STOPSIGNAL SIGKILL
+
 # Do the streamer ready-check and start
 CMD /app/start.sh
--- a/src/docker/processor/Dockerfile
+++ b/src/docker/processor/Dockerfile
@@ -50,8 +50,6 @@ RUN rm g.tar.gz
 
 WORKDIR /app
 COPY /src/docker/processor/start.sh .
-RUN mkdir -p /config/data
-COPY /src/docker/processor/config.yaml /config/data/config.yaml
 RUN chmod u+x /app/start.sh
 WORKDIR /
 
--- a/src/docker/processor/Dockerfile
+++ b/src/docker/processor/Dockerfile
@@ -39,6 +39,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         net-tools \
         netcat \
         tcpdump \
+        postgresql-client \
     && rm -rf /var/lib/apt/lists/*
 COPY --from=builder $BIN_PATH /usr/local/bin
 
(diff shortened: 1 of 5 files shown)


alnoki:
Per offline discussion

Rather than mounting in the config, how about?

 - [ ] Use [docker environment file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file) for all environment variables, that user might set, including Aptos Network and modifiable processor config options
- [ ] Include aptos_network in the environment file as well
- [ ] Have processor startup script generate a config.yaml file from environment variables
- [ ] Only add the "starting version" field to the processor's onboard config if db is empty for given processor status row
- [ ] Update docker readme
- [ ] Update DSS homepage instructions for new env file
- [ ] Merge this to `dss-v1.5.0` branch so that docs changes don't land before stable build is ready

src/docker/processor/Dockerfile
```diff
@@ -49,6 +50,8 @@ RUN rm g.tar.gz
 
 WORKDIR /app
 COPY /src/docker/processor/start.sh .
+RUN mkdir -p /config/data
+COPY /src/docker/processor/config.yaml /config/data/config.yaml
```
alnoki, line 54: Is this necessary?

src/docker/processor/start.sh
```diff
@@ -13,4 +13,23 @@ if [[ "$HEALTHCHECK_BEFORE_START" == "true" ]];then
     done
 fi
 
-/usr/local/bin/processor -c /config/data/config.yaml
+psql $DATABASE_URL -c '\copy processor_status to out.csv csv'
+
+if [ -s "out.csv" ];then
+    export STARTING_VERSION=$(cut -d, -f2 out.csv)
+fi
```
alnoki, line 20: When trying to deploy on cloud I've noticed issues when storing in root dir

How about storing at a `/app/status_at_startup.csv`?

alnoki (changes requested):
Given that #658 will require the new config style, it probably makes sense to

1. Address comments
2. Get this merged to `v1.5.0`
3. Merge `v1.5.0` into #658
4. Merge #658 into `v1.5.0`
5. Add a docs PR into `v1.5.0` as described at https://github.com/econia-labs/econia/pull/659#issuecomment-1858420925 (I've created ECO-1080 to track this)

alnoki (approved):

src/docker/processor/start.sh
```diff
@@ -13,4 +13,25 @@ if [[ "$HEALTHCHECK_BEFORE_START" == "true" ]];then
     done
 fi
 
-/usr/local/bin/processor -c /config/data/config.yaml
+psql $DATABASE_URL -c '\copy processor_status to /app/out.csv csv'
+
+if [ -s "out.csv" ];then
+    export STARTING_VERSION=$(cut -d, -f2 /app/out.csv)
```
alnoki, line 19: This should be `[ -s "/app/out.csv" ]`

Will fix in #657

alnoki: Fixed at https://github.com/econia-labs/econia/pull/657/commits/b3a36d4ae04e2fd18ad36e110da46d8b58903075

PR #662, opened 2023-12-20
[ECO-1086] add a script to init grafana

CRBl69:
This PR adds a script that creates a PostgreSQL data source for Grafana, as well as a new public user.

It needs the following environment variables:

- ADMIN_PASSWORD (ex: 1234)
- GRAFANA_HOST (ex: grafana.econia.lol)
- GRAFANA_PORT (ex: 3000)
- DATABASE_HOST (ex: database.econia.lol)
- DATABASE_PORT (ex: 5432)
- DATABASE_USER (ex: grafana)
- DATABASE_PASSWORD (ex: *****)
- DATABASE_DB (ex: econia)
- PUBLIC_USER_EMAIL (ex: mister.potato1984@example.com)
- PUBLIC_USER_PASSWORD (ex: *****)

### DIFF
--- a/src/terraform/internal-dss/create_grafana_user.sh
+++ b/src/terraform/internal-dss/create_grafana_user.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# Create a PostgreSQL data source
+curl "http://admin:$ADMIN_PASSWORD@$GRAFANA_HOST:$GRAFANA_PORT/api/datasources" \
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    --data "{\"name\":\"DSS Database\",\"type\":\"postgres\",\"url\":\"$DATABASE_HOST:$DATABASE_PORT\",\"access\":\"proxy\",\"user\":\"$DATABASE_USER\",\"password\":\"$DATABASE_PASSWORD\",\"jsonData\": {\"database\":\"$DATABASE_DB\",\"sslmode\":\"disable\",\"postgresVersion\":1500}}"
+
+# Create a public user
+curl "http://admin:$ADMIN_PASSWORD@$GRAFANA_HOST:$GRAFANA_PORT/api/admin/users" \
+    -X POST \
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    --data "{\"name\":\"public\",\"email\":\"$PUBLIC_USER_EMAIL\",\"login\":\"public\",\"password\":\"$PUBLIC_USER_PASSWORD\",\"OrgId\":1}"


src/terraform/internal-dss/create_grafana_user.sh
```diff
@@ -0,0 +1,14 @@
...
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    --data "{\"name\":\"DSS Database\",\"type\":\"postgres\",\"url\":\"$DATABASE_HOST:$DATABASE_PORT\",\"access\":\"proxy\",\"user\":\"$DATABASE_USER\",\"password\":\"$DATABASE_PASSWORD\",\"jsonData\": {\"database\":\"$DATABASE_DB\",\"sslmode\":\"disable\",\"postgresVersion\":1500}}"
+
+# Create a public user
+curl "http://admin:$ADMIN_PASSWORD@$GRAFANA_HOST:$GRAFANA_PORT/api/admin/users" \
+    -X POST \
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    --data "{\"name\":\"public\",\"email\":\"$PUBLIC_USER_EMAIL\",\"login\":\"public\",\"password\":\"$PUBLIC_USER_PASSWORD\",\"OrgId\":1}"
```
alnoki, line 14: How about changing this to use `public` as the username instead of having a public user email? That will make it easier to log in using `public` username

src/terraform/internal-dss/create_grafana_user.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# Create a PostgreSQL data source
+curl "http://admin:$ADMIN_PASSWORD@$GRAFANA_HOST:$GRAFANA_PORT/api/datasources" \
+    -H 'Accept: application/json' \
+    -H 'Content-Type: application/json' \
+    --data "{\"name\":\"DSS Database\",\"type\":\"postgres\",\"url\":\"$DATABASE_HOST:$DATABASE_PORT\",\"access\":\"proxy\",\"user\":\"$DATABASE_USER\",\"password\":\"$DATABASE_PASSWORD\",\"jsonData\": {\"database\":\"$DATABASE_DB\",\"sslmode\":\"disable\",\"postgresVersion\":1500}}"
```
alnoki, line 7: Does this correspond to postgres v14? That is what GCP is using

alnoki (changes requested):

src/terraform/internal-dss/create_grafana_user.sh
alnoki: Did you test this out? It is quite different from what ended up being required on GCP:

https://github.com/econia-labs/econia/blob/af85fe5131ceea4ac879637c157f682621eabd66/src/terraform/dss-ci-cd/dss/modules/grafana/main.tf#L66-L94

src/terraform/internal-dss/create_grafana_user.sh
alnoki: Note org ID isn't necessary:

https://github.com/econia-labs/econia/blob/af85fe5131ceea4ac879637c157f682621eabd66/src/terraform/dss-ci-cd/dss/modules/grafana/main.tf#L114-L130

alnoki (changes requested):
How about?

- [ ] Moving this into `src/docker/grafana`
- [ ] Add a `src/docker/grafana/README` that shows starting local grafana once you already have a DSS
- [ ] Add to the init scripts a call to insert the annotations data from #667
- [ ]  Use #667 as a base branch
- [ ]  If possible, make the JSON part of the commands more readable, e.g.

```sh
"{\
    \"field\": \"$value"\ 
...
}"
```

etc.

src/terraform/internal-dss/create_grafana_user.sh
alnoki: > The version does not seem to be important

Not now, at least, but if a PostgreSQL version major gets released then perhaps the new defaults could introduce unexpected behaviors.

Our postgres DSS db Dockerfile pins to v14, as does our GCP deployment for both DSS db and Grafana state, so I'm thinking we should probably create a new Postgres docker compose container for v14 to store Grafana state: this will make it easier to migrate grafana state back and forth between cloud and local, since Grafana defaults to SQLite for storing state

PR #667, opened 2023-12-22
[ECO-1102] add annotations for grafana

CRBl69:
This PR adds annotations. It's a table that will allow displaying meaningful events in the grafana views.


src/rust/dbv2/migrations/2023-12-21-132732_annotations/up.sql
```diff
@@ -0,0 +1,32 @@
...
+    "title" text NOT NULL,
+    "text" text,
+    "tag" text NOT NULL,
+);
+
+INSERT INTO annotations ("time", "title", "tag") VALUES (
+    '2023-11-28 14:03:00+00'::timestamptz,
+    'Econia mainnet launch',
+    'Econia'
+);
```
alnoki, line 14: I don't think migrations is the right place for the actual dataset

What about a separate json file?

Then we can do a CI/CD script that just loads the data from a git commit, etc.

Maybe the JSON file could be at `/src/rust/dbv2/grafana-annotations.json?`

src/docker/compose.dss-core.yaml
```diff
@@ -19,6 +19,20 @@ services:
       APTOS_NETWORK: ${APTOS_NETWORK}
       DATABASE_URL: "postgres://econia:econia@postgres:5432/econia"
 
+  grafana:
```
alnoki, line 22: How about calling this something like `grafana_annotations`?

That way we can have another service that also just runs the Grafana image

src/rust/dbv2/migrations/2023-12-21-132732_annotations/up.sql
```diff
@@ -0,0 +1,9 @@
+-- Your SQL goes here
+CREATE TABLE annotations (
```
alnoki, line 2: Would it make more sense to have this be under a schema `grafana.annotations`?

Or maybe just a different table name like `grafana_event_annotations` or even `grafana_annotations`?

Also, can we add some kind of spec on the table definition that skips if trying to re-insert the same events (for when a new annotation gets added)?

alnoki:
Per offline discussion, standing by for testing completion

alnoki (approved):

PR #678, opened 2024-01-29
[ECO-1153] add more documentation before publishing the rust crate

CRBl69:
Add some more Rust doc before publishing the crate to make reading the doc and using the crate more intuitive.

### DIFF
--- a/src/rust/sdk/README.md
+++ b/src/rust/sdk/README.md
@@ -1,4 +1,4 @@
 # Econia SDK
 
 The Econia SDK is documented using Rust comments.
-You can check it out on the [docs.rs](https://docs.rs/) website, or locally, by running `cargo doc` and then opening the generated files (the one which's path is printed during doc compilation).
+You can check it out on the [docs.rs](https://docs.rs/) website, or locally, by running `cargo doc --open`.
--- a/src/rust/sdk/README.md
+++ b/src/rust/sdk/README.md
@@ -1,0 +1,4 @@
+# Econia SDK
+
+The Econia SDK is documented using Rust comments.
+You can check it out on the [docs.rs](https://docs.rs/) website, or locally, by running `cargo doc` and then opening the generated files (the one which's path is printed during doc compilation).
--- a/src/rust/sdk/src/lib.rs
+++ b/src/rust/sdk/src/lib.rs
@@ -1,3 +1,16 @@
+//! # Econia SDK
+//!
+//! The Econia Rust SDK is an SDK to interact with the Econia V4 protocol. You can read more about
+//! the protocol at <https://econia.dev/>.
+//!
+//! The workflow to use this SDK is quite straight forward (only 3 steps):
+//!
+//! 1. Create a [`EconiaClient`]
+//! 2.
+//!     - Either use it to get a [`EconiaViewClient`]
+//!     - Or create a payload using helper functions from the [`entry`] module
+//! 3. There is no 3rd step.
+
 use aptos_api_types::{
     AptosErrorCode, MoveType, Transaction, TransactionInfo, UserTransactionRequest, VersionedEvent,
     U64,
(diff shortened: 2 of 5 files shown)


alnoki (changes requested):
How about a minor README on how to generate the documentation and view?

src/rust/sdk/src/lib.rs
```diff
@@ -1,3 +1,16 @@
...
+//! The Econia Rust SDK is an SDK to interact with the Econia V4 protocol. You can read more about
+//! the protocol at <https://econia.dev/>.
+//!
+//! The workflow to use this SDK is quite straight forward (only 3 steps):
+//!
+//! 1. Create a [`EconiaClient`]
+//! 2.
+//!     - Either use it to get a [`EconiaViewClient`]
+//!     - Or create a payload using helper functions from the [`entry`] module
+//! 3. There is no 3rd step.
```
alnoki, line 12: How about taking this out?

src/rust/sdk/src/lib.rs
```diff
@@ -1,3 +1,16 @@
...
+//!
+//! The Econia Rust SDK is an SDK to interact with the Econia V4 protocol. You can read more about
+//! the protocol at <https://econia.dev/>.
+//!
+//! The workflow to use this SDK is quite straight forward (only 3 steps):
+//!
+//! 1. Create a [`EconiaClient`]
+//! 2.
+//!     - Either use it to get a [`EconiaViewClient`]
+//!     - Or create a payload using helper functions from the [`entry`] module
```
alnoki, line 11: How about something like?

"To use the SDK create an [`EconiaClient`], then either use it to get an [`EconiaViewClient`], or create a payload using helper functions from the [`entry`] module"

src/rust/sdk/src/view.rs
```diff
@@ -1,3 +1,13 @@
+//! The [`view`](crate::view) module exposes the view functionalities of the Econia package.
+//!
+//! You can create a [`EconiaViewClient`] from a [`EconiaClient`](crate::EconiaClient) using the
```
alnoki, line 3: "a" -> "an" in both cases in this line

src/rust/sdk/src/view.rs
```diff
@@ -1,3 +1,13 @@
...
+//! [`EconiaClient::view_client`](crate::EconiaClient::view_client) function.
+//!
+//! You can then call the functions provided here.
+//!
+//! There are also stand alone functions which do not need any blockchain state and compute
```
alnoki, line 8: stand alone -> standalone

src/rust/sdk/src/view.rs
```diff
@@ -1,3 +1,13 @@
...
+//! locally. Those do not need a [`EconiaViewClient`] to execute.
```
alnoki, line 9: "a" -> "an"

alnoki (changes requested):

src/rust/sdk/README.md
alnoki, line 1: Admittedly I'm not too familiar with the `cargo doc` process, but still, I notice that when I run `cargo doc --open` the "crates" column has a ton of ineherited crates along the left side that are confusing in the context of the SDK

Is there a way to resolve this to point at an Aptos crate? Or perhaps to make the dependencies show up somewhere other than the `Crates` column navigator? Because I have to scroll down multiple pages to find the `econia_sdk` page

alnoki (changes requested):

src/rust/sdk/README.md
CRBl69: I see what you mean, but I don't see how that is a problem since you are already on the `econia_sdk` page whenever you use `--open` or you click on the printed file path. And this won't be an issue when accessing [docs.rs](https://docs.rs/).

src/rust/sdk/README.md
```diff
@@ -0,0 +1,4 @@
+# Econia SDK
+
+The Econia SDK is documented using Rust comments.
+You can check it out on the [docs.rs](https://docs.rs/) website, or locally, by running `cargo doc` and then opening the generated files (the one which's path is printed during doc compilation).
```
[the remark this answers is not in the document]
CRBl69: I see what you mean. Just FYI, this is what I see.

image

alnoki (approved):

PR #680, opened 2024-01-30
[ECO-1170] return market volume data in subunits

CRBl69:
This PR makes it so that when querying the `/markets` endpoint, we return volume in subunits rather than ticks so that it is easier for the endpoint users to use this data.

### DIFF
--- a/src/rust/dbv2/migrations/2024-01-30-162544_markets_vol_in_subunits/up.sql
+++ b/src/rust/dbv2/migrations/2024-01-30-162544_markets_vol_in_subunits/up.sql
@@ -1,0 +1,105 @@
+-- Your SQL goes here
+CREATE MATERIALIZED VIEW aggregator.markets_24h_data2 AS
+    WITH fills AS(
+        SELECT
+            *
+        FROM
+            fill_events
+        WHERE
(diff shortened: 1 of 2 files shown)


src/rust/dbv2/migrations/2024-01-30-162544_markets_vol_in_subunits/up.sql
alnoki, line 1: There are enough dependencies on the DSS at this point that I hesitate to approve changes changing the format of an existing field. For example our reference frontend would break with these changes, as would other trading frontends.

Instead, how about leaving all pre-existing fields as they were, and adding as needed?

- [ ] `cumulative_quote_volume_in_subunits`
- [ ] `base_volume_24h_in_subunits`
- [ ] `quote_volume_24h_in_subunits`

alnoki (changes requested):

src/rust/dbv2/migrations/2024-01-30-162544_markets_vol_in_subunits/up.sql
alnoki: Noting per offline discussion we agreed this is not required for Defi LLama integration

alnoki:
Closing since not needed for DeFi LLama integration

PR #689, opened 2024-02-04
[ECO-1222] make aggregator process certain data in batch to avoid oom kills

CRBl69:
We experienced aggregator crashes due to out of memory errors.

This PR batches `SELECT` statements which load too much data into RAM at once into smaller, parametrable batches, which avoid this problem.


alnoki:
I was unable to cleanly change this PR's base branch to the `testing` branch from #688 without introducing merge conflicts, so I checked out #688 and manually merged this in

alnoki (approved):

PR #691, opened 2024-02-05
[ECO-1225] Publicly expose select private view functions

alnoki:
Presently, select view functions are kept private to prevent runtime state contention for order books, which are global resources within a market. However with Aptos block space unsaturated these view functions can be publicly exposed without the expectation of significant performance disadvantages, enabling greater composability.

Since the Econia Move package is already close to the 64kb Aptos txn boundary, this PR only captures functions that can be easily exposed via the addition of the `public` keyword to their function signatures, so as to reduce bytecode bloat.

Publicly expose the following functions:

1. `market::get_open_orders_paginated()`
1. `market::get_price_levels_paginated()`
1. `market::has_open_order()`
1. `market::get_open_order()`

Add the following functions to decode view function returns:

1. `market::get_order_view_fields()`
2. `market::get_orders_view_fields()`
3. `market::get_price_level_fields()`
4. `market::get_price_levels_fields()`

# Audit changeset commentary

* Only `market.move` source needs to be reviewed, the remaining changeset is just docs and tests

# Testing

* Tested to 100% coverage using 50574ba
* Verified backwards compatible with Econia https://github.com/econia-labs/econia/releases/tag/v4.2.1-audited via v2 multisig publication on devnet


xbtmatt (approved):
Focused mostly on the `market.move` changes, pretty straightforward, looks good!

PR #693, opened 2024-02-08
[ECO-1183] add fees pipeline

CRBl69:
This PR adds a fees pipeline. This will allow more in depth visualizations in Grafana and integration with DefiLlama.

# Example

```http
GET /fees?market_id=eq.7&time=eq.2023-12-20%2022:00:00
```

```json
[
  {
    "time": "2023-12-20T22:00:00+00:00",
    "market_id": 7,
    "amount": 3650202
  }
]
```

```http
GET /rpc/get_market_cumulative_fees?market_id=7&up_to=eq.2023-12-20%2022:00:00
```

```
3030024956
```


src/rust/dbv2/migrations/2024-02-07-144601_fees/up.sql
```diff
@@ -0,0 +1,33 @@
+-- Your SQL goes here
+CREATE TABLE aggregator.fees (
+    "time" TIMESTAMPTZ,
```
alnoki, line 3: It looks like we are calculating fees over the course of an hour, but that might not be immediately clear to someone querying this table from PostgREST

What about `"start_time_1hr_period"`?

src/rust/aggregator/src/pipelines/fees.rs
```diff
@@ -0,0 +1,60 @@
...
+        let mut transaction = create_repeatable_read_transaction(&self.pool).await?;
+        sqlx::query_file!("sqlx_queries/fees/backfill.sql",)
+            .execute(&mut transaction as &mut PgConnection)
+            .await
+            .map_err(to_pipeline_error)?;
+        sqlx::query_file!("sqlx_queries/fees/delete_last_indexed_txn.sql",)
+            .execute(&mut transaction as &mut PgConnection)
+            .await
+            .map_err(to_pipeline_error)?;
+        sqlx::query_file!("sqlx_queries/fees/update_last_indexed_txn.sql",)
```
alnoki, line 53: How about just using a simple SQL statement to cover these two? E.g. one that tries to insert, and overwrites on conflict?

src/rust/dbv2/migrations/2024-02-07-144601_fees/up.sql
```diff
@@ -0,0 +1,33 @@
...
+    "market_id" NUMERIC(20,0),
+    "amount" NUMERIC(39,0),
```
alnoki, line 5: How about `"fees_in_quote_subunits"`, since plenty of DEXes, e.g. CPMMs charge in both base and quote?

alnoki (changes requested):

src/rust/dbv2/migrations/2024-02-07-144601_fees/up.sql
CRBl69: :+1:

src/rust/aggregator/src/pipelines/fees.rs
alnoki: I may have miscommunicated above, because I'm confused by your response. Perhaps what I mean was:

How about combining `delete_last_indexed_txn.sql` and `update_last_indexed_txn.sql` into a single query?

Not critical though, and the functionality is still assured by the current logic so I'll mark this as resolved

src/rust/aggregator/sqlx_queries/fees/backfill.sql
alnoki, line 1: `cargo check` is failing due to this file. Did you recreate the SQLx queries?

alnoki (changes requested):

src/rust/aggregator/src/pipelines/fees.rs
alnoki: Per offline discussion, noting that the intended consolidation could be achieved via an `UPDATE` statement per `user_history/set_new_last_indexed_txn_version.sql`, but is not required

alnoki (approved):

PR #699, opened 2024-02-13
Add restart option and postgrest max row environment

harrier999:
To resolve a few issues that arose while operating the node, added several options to the docker-compose file.

1. Add option `restart: unless-stopped` to 
     1. aggregator
     2. postgrest
     3.  ws
     4. processor

2. Add enviroment value `PGRST_DB_MAX_ROWS:500` to postgrest

### DIFF
--- a/src/docker/compose.dss-core.yaml
+++ b/src/docker/compose.dss-core.yaml
@@ -45,7 +45,7 @@ services:
       PGRST_DB_URI: "postgres://econia:econia@postgres:5432/econia"
       PGRST_DB_ANON_ROLE: web_anon
       PGRST_DB_SCHEMA: api
-      PGRST_DB_MAX_ROWS: 500
+      PGRST_DB_MAX_ROWS: ${POSTGREST_MAX_ROWS}
     image: postgrest/postgrest
     ports:
       - "3000:3000"
--- a/src/docker/compose.dss-core.yaml
+++ b/src/docker/compose.dss-core.yaml
@@ -18,6 +18,8 @@ services:
     environment:
       APTOS_NETWORK: ${APTOS_NETWORK}
       DATABASE_URL: "postgres://econia:econia@postgres:5432/econia"
+    restart: unless-stopped
+
 
   grafana_annotations:
     build:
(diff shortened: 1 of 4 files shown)


CRBl69:
Hi. Thank you for the PR. I think the `restart` flag is a good idea :+1:

As for `PGRST_DB_MAX_ROWS`, you about putting this into an env variable called `POSTGREST_MAX_ROWS` ? This could then be set in the `.env` file, along side other env variables.

src/docker/compose.dss-global.yaml
```diff
@@ -13,3 +13,4 @@ services:
       extends:
         file: compose.processor-template.yaml
         service: processor-template
+      restart: unless-stopped
```
alnoki, line 16: Will you please move this to `compose.processor-template.yaml`?

That way it will be used for both a global DSS (on a public chain) and for a local DSS (on a local testnet)

src/docker/compose.dss-core.yaml
```diff
@@ -43,9 +45,11 @@ services:
       PGRST_DB_URI: "postgres://econia:econia@postgres:5432/econia"
       PGRST_DB_ANON_ROLE: web_anon
       PGRST_DB_SCHEMA: api
+      PGRST_DB_MAX_ROWS: 500
```
alnoki, line 48: How about adding to `example.env` a variable `POSTGREST_MAX_ROWS`?

Then, in this file you can use `PGRST_DB_MAX_ROWS: ${POSTGREST_MAX_ROWS}`

That will allow DSS users to set all configs in their local copy of `example.env`

alnoki (changes requested):
@harrier999 thank you for your contribution!

I've requested two minor changes that should make for an easier UX

src/docker/compose.dss-global.yaml
harrier999: Thank you for your feedback, I moved it to `compose.processor-template.yaml`

harrier999:
Thank you for your feedback, everyone!
I have incorporated your feedback. Could you please review it?

CRBl69 (approved):

alnoki (approved):

alnoki:
@harrier999 thank you for your contribution and for a prompt resolution to review comments!

PR #700, opened 2024-02-15
[ECO-1280] add spreads

CRBl69:
This PR adds a new table that can be used to get min ask, max bid, spread and mid price.

# Example

```http
GET /spreads
```

```json
[
  {
    "market_id": 7,
    "time": "2023-12-26T08:11:00+00:00",
    "min_ask": 10990,
    "max_bid": 9510
  }
]
```

### DIFF
--- a/src/rust/.sqlx/query-6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de.json
+++ b/src/rust/.sqlx/query-6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de.json
@@ -1,0 +1,17 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "WITH parameters AS (\n    SELECT\n        $1::numeric(20,0) AS \"market_id\",\n        $2::timestamptz AS \"time\",\n        $3::numeric(20,0) AS min_ask,\n        $4::numeric(20,0) AS max_bid\n)\nINSERT INTO aggregator.spreads\nSELECT\n    *\nFROM\n    parameters;\n",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Numeric",
+        "Timestamptz",
+        "Numeric",
+        "Numeric"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de"
+}
--- a/src/rust/aggregator/src/pipelines/spreads.rs
+++ b/src/rust/aggregator/src/pipelines/spreads.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 
+use anyhow::bail;
 use bigdecimal::{BigDecimal, Zero};
 use chrono::{DateTime, Duration, Utc};
 use sqlx::{PgConnection, PgPool, Transaction};
@@ -10,6 +11,8 @@ use sqlx_postgres::Postgres;
 use crate::MAX_BATCH_SIZE;
 
 pub const TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
+
+const ASK: bool = true;
 
 pub struct Spreads {
     pool: PgPool,
@@ -27,21 +30,14 @@ impl Spreads {
     }
 }
 
+// The first element is market_id, the second one is order_id.
+// This allows us to uniquely identify an order.
 type OrderKey = (BigDecimal, BigDecimal);
 
 #[derive(Clone)]
 struct SpreadsState {
     orders: HashMap<OrderKey, Order>,
     last_txn_indexed: BigDecimal,
-}
-
-impl SpreadsState {
-    pub fn new(last_txn: BigDecimal, orders: HashMap<OrderKey, Order>) -> Self {
-        Self {
-            last_txn_indexed: last_txn,
-            orders,
-        }
-    }
 }
 
 impl Default for SpreadsState {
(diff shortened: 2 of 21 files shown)


src/rust/aggregator/src/pipelines/spreads.rs
```diff
@@ -0,0 +1,373 @@
...
+        };
+        let intervals = (latest_event_timestamp - last_indexed_timestamp).num_minutes() as usize;
+        let mut timestamps = Vec::with_capacity(intervals);
+        for i in 0..intervals {
+            timestamps.push(last_indexed_timestamp + chrono::Duration::minutes(i as i64 + 1));
+        }
+        if timestamps.len() == 0 {
+            return Ok(());
+        }
+        let mut state = self.state.clone().unwrap_or_default();
```
alnoki, line 324: Is there anyway we can mutably borrow this instead of cloning? Pass-by-value here could get expensive

src/rust/aggregator/src/pipelines/spreads.rs
```diff
@@ -0,0 +1,373 @@
...
+                    (None, Some(change)) => {
+                        handle_change(change, &mut orders, &mut changes_index);
+                    }
+                    _ => unreachable!(),
+                };
+            }
+
+            txn_version_of_state = txn_version;
+        }
+    }
```
alnoki, line 231: No blank line here?

Should probably run `rustfmt`

src/rust/aggregator/src/pipelines/spreads.rs
```diff
@@ -0,0 +1,373 @@
...
+    fn get_spreads_form_state<'a>(
+        &self,
+        orders: &HashMap<OrderKey, Order>,
+    ) -> (HashMap<BigDecimal, BigDecimal>, HashMap<BigDecimal, BigDecimal>) {
+
+        let mut min_asks: HashMap<BigDecimal, BigDecimal> = Default::default();
+        let mut max_bids: HashMap<BigDecimal, BigDecimal> = Default::default();
+
+        for element in orders.values().cloned() {
```
alnoki, line 240: Avoid cloning if possible, there's a decent amount of state here

src/rust/aggregator/src/pipelines/spreads.rs
```diff
@@ -0,0 +1,373 @@
...
+            if element.side {
```
alnoki, line 241: How about pinning this to something explicit, like `element.side == ASK`?

src/rust/.sqlx/query-6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de.json
alnoki, line 1: I count 6 SQLx files but many more SQL queries than that: maybe re-run the SQLx script?

alnoki (changes requested):

src/rust/aggregator/src/pipelines/spreads.rs
```diff
@@ -0,0 +1,373 @@
...
+            // Handle fills and changes chronologically.
+            let mut fills_index = 0;
+            let mut changes_index = 0;
+            for _ in 0..(fills.len() + changes.len()) {
+                let fill = fills.get(fills_index);
+                let change = changes.get(changes_index);
+                let (f, c) = match (fill, change) {
+                    (Some(fill), Some(change)) => {
+                        if (fill.txn_version.clone(), fill.event_idx.clone())
+                            < (change.txn_version.clone(), change.event_idx.clone())
```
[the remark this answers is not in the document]
CRBl69: Yes. Here is an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=14a27e1dec5734e3232b24d3252f07ab.

src/rust/aggregator/src/pipelines/spreads.rs
CRBl69: There is some cloning that needs to be done anyways.

It's either that, or:

```rust
for element in state.orders.values() {
    if element.side {
        if let Some(min) = min_asks.get(&element.market_id) {
            if *min > element.price {
                min_asks.insert(element.market_id.clone(), element.price.clone());
            }
        } else {
            min_asks.insert(element.market_id.clone(), element.price.clone());
        }
    } else {
        if let Some(max) = max_bids.get(&element.market_id) {
            if *max < element.price {
                max_bids.insert(element.market_id.clone(), element.price.clone());
            }
        } else {
            max_bids.insert(element.market_id.clone(), element.price.clone());
        }
    }
}
```

Even though in the second version, we only clone `price` and `market_id` instead of the entire order struct, I'm wondering if due to the way cloning works, it isn't more optimized to clone the whole array at once.

src/rust/.sqlx/query-6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de.json
CRBl69: This may be because some are copy pasted from the `order_histoy` pipeline. Those will be gone once we replace all the metrics we need with the more optimized pipelines.

src/rust/aggregator/src/pipelines/spreads.rs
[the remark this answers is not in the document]
alnoki: Excellent, thank you for the clear example

src/rust/aggregator/src/pipelines/spreads.rs
alnoki: Cloning `price` and `market_id` only is manageable, and the clone only happens when a new min/max is found so only some of the data will get cloned. This new approach seems more efficient

src/rust/.sqlx/query-6d0ae6fd5f1dca3a60979fcf01611c95a5d2ea96678841c51fd04de89bf757de.json
alnoki: Roger, noting I just ran the SQLx generator and migrations checker scripts without issue

src/rust/aggregator/src/pipelines/spreads.rs
alnoki: Strangely enough my format on save for VS code doesn't even modify this

Noting that we'll probably want to have a strict Rust formatter/linter at some point to avoid such issues

Bookmarking this comment internally for the CI/CD work we have scoped out

alnoki (approved):

PR #724, opened 2024-03-19
[ECO-1445] convert u128 to strings for serialization

CRBl69:
This is needed so that it is easy to deserialize JSONs in JavaScript.

Long story short, JSON supports infinity large numbers, but the JavaScript implementation of the JSON parser doesn't (ironic, I know). Thus, we convert the fields to strings before serializing, so that JavaScript doesn't round up numbers on deserialization.

The world if every implementation correcty implemented the spec:

image


alnoki:
Per eng call, we decided to close this because the order IDs are emitted correctly at the MQTT level. It is not the DSS's responsibility to defensively guard against serialization issues from external libraries. If an external library has a known serialization issue it should be addressed externally

PR #729, opened 2024-03-20
[ECO-1338] optimize tickers

CRBl69:
This PR brings down execution time of tickers from seconds to 400ms.


src/rust/dbv2/migrations/2024-03-18-130309_optimize_tickers/up.sql
```diff
@@ -0,0 +1,32 @@
...
+    SELECT max_bid
+    FROM aggregator.spreads
+    WHERE market_id = $1
+    ORDER BY "time" DESC
+    LIMIT 1
+$$ LANGUAGE SQL;
+
+
+CREATE OR REPLACE FUNCTION api.get_market_liquidity (market_id numeric, depth numeric) RETURNS NUMERIC AS $$
+    SELECT amount_ask_lots + amount_bid_lots FROM aggregator.liquidity WHERE group_id = (SELECT group_id FROM aggregator.liquidity_groups WHERE name = 'all' AND market_id = $1) AND bps_times_ten = $2 * 10 ORDER BY "time" DESC LIMIT 1;
```
alnoki, line 31: It looks like this is reporting liquidity in lots

Why is there no conversion to ticks or to quote indivisible subunits?

alnoki:
- [ ] Noting that we should merge this after refactoring liquidity calculations

src/rust/dbv2/migrations/2024-03-21-183256_reset_liquidity/down.sql
alnoki, line 1: Now that #730 has merged I think you'll need to merge `main` back in

Because, for example, if this migration merges to main it'll revert the fix requested here: https://github.com/econia-labs/econia/pull/730#discussion_r1534789538

And for good measure, once that's done:

- [ ] Run SQLx generator
- [ ] Run migrations checker

alnoki (changes requested):

PR #730, opened 2024-03-21
[ECO-1450] fix liquidity calculation logic

CRBl69:
This PR fixes the calculation of liquidity.

Note that we drop data in the order history pipelines tables to trigger recalculation.


src/rust/aggregator/src/pipelines/order_history_pipelines.rs
```diff
@@ -352,15 +352,15 @@ fn add_to_map<'a>(
         map.entry(LiquidityKey { group_id, bps_times_tem: bps_times_ten })
             .and_modify(|e| e.base += &order.size * &order.price)
             .or_insert(LiquidityValue {
-                base: &order.size * &order.price,
+                base: &order.size * price,
```
alnoki, line 355: I traced this logic back up the stack and see that it is coming from `get_cached_last_price`

Does this mean that during backfilling, the price used will be current last price, or the last price at the relevant time?

src/rust/aggregator/src/pipelines/order_history_pipelines.rs
alnoki, line 1: Since you're updating this file, a few suggestions to sneak in:

```suggestion
#[derive(PartialEq, Eq, Hash)]
struct LiquidityKey {
    group_id: i32,
    bps_times_ten: i32,
}
```

(`times_tem` to `times_ten`)


```suggestion
struct LiquidityValue {
    /// Effective value of base locked in asks, quoted in ticks.
    /// Price taken as the last fill price at the time of calculation.
    base: BigDecimal,
    /// Amount of quote locked in bids, quoted in ticks.
    quote: BigDecimal,
}
```

(Clarifying comments)

src/rust/dbv2/migrations/2024-03-21-183256_reset_liquidity/down.sql
```diff
@@ -0,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+DELETE FROM aggregator.liquidity;
+DELETE FROM aggregator.spreads;
+DELETE FROM aggregator.order_history_last_indexed_timestamp;
```
alnoki, line 4: Is this missing column modification?

alnoki (changes requested):

src/rust/aggregator/src/pipelines/order_history_pipelines.rs
CRBl69: It will always use the price at the relevant time. This is done so that you don't query the last price from the DB every time you need it. Since you might have 500 orders and one last price, there is no need to query the last price 500 times.

alnoki (approved):

PR #731, opened 2024-03-21
[ECO-1452] Add hot upgrade docs support

alnoki:
1. Add a hot upgrade section to the DSS landing page
2. Annotate v2.0.1 as hot upgradable
3. Update release procedure with hot upgrade step

### DIFF
--- a/doc/doc-site/docs/off-chain/dss/data-service-stack.md
+++ b/doc/doc-site/docs/off-chain/dss/data-service-stack.md
@@ -62,6 +62,7 @@ Before you start working with the DSS, make sure you are on the right branch and
 ```bash
 # From Econia repo root
 git checkout dss-stable
+git pull
 git submodule update --init --recursive
 ```
 
@@ -129,7 +130,6 @@ See the [changelog](./changelog.md) to verify that the latest release supports a
 1. Clean your Docker containers and image cache:
 
    ```bash
-   docker ps -aq | xargs docker stop | xargs docker rm
    docker system prune -af
    ```
 
@@ -141,13 +141,14 @@ See the [changelog](./changelog.md) to verify that the latest release supports a
 
    ```bash
    git checkout dss-stable
+   git pull
    git submodule update --init --recursive
    ```
 
 1. Restart the DSS:
 
    ```bash
-   docker compose --file src/docker/compose.dss-core.yaml up
+   docker compose --file src/docker/compose.dss-core.yaml up --force-recreate
    ```
 
 ### Verifying the DSS
--- a/doc/doc-site/docs/off-chain/dss/data-service-stack.md
+++ b/doc/doc-site/docs/off-chain/dss/data-service-stack.md
@@ -109,6 +109,47 @@ Finally, to fully shut it down:
 docker compose --file src/docker/compose.dss-core.yaml down
 ```
 
+### Conducting a hot upgrade
+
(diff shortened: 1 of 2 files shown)


doc/doc-site/docs/off-chain/dss/data-service-stack.md
```diff
@@ -109,6 +109,47 @@ Finally, to fully shut it down:
...
+   ```
+
+   :::tip
+   This is similar to starting a DSS from scratch, but it preserves the `db` volume.
+   :::
+
+1. Check out the latest stable DSS source code:
+
+   ```bash
+   git checkout dss-stable
```
CRBl69, line 143: How about adding a `git pull` instruction ?

doc/doc-site/docs/off-chain/dss/data-service-stack.md
```diff
@@ -109,6 +109,47 @@ Finally, to fully shut it down:
...
+1. Shut down the DSS
+
+   ```bash
+   docker compose --file src/docker/compose.dss-core.yaml down
+   ```
+
+1. Clean your Docker containers and image cache:
+
+   ```bash
+   docker ps -aq | xargs docker stop | xargs docker rm
```
CRBl69, line 132: I do not think the `docker stop` part is necessary, you just `down`ed everything.

CRBl69: Also, since you just `down`ed everything, `ps` doesn't return anything (just checked that).

doc/doc-site/docs/off-chain/dss/data-service-stack.md
```diff
@@ -109,6 +109,47 @@ Finally, to fully shut it down:
...
+
+   ```bash
+   git checkout dss-stable
+   git submodule update --init --recursive
+   ```
+
+1. Restart the DSS:
+
+   ```bash
+   docker compose --file src/docker/compose.dss-core.yaml up
```
CRBl69, line 150: How about adding a `docker compose --file src/docker/compose.dss-core.yaml build --no-cache` to make sure everything is correctly rebuilt. In my experience, Docker often doesn't realize something has changed and does not correctly rebuild images.

CRBl69 (changes requested):

doc/doc-site/docs/off-chain/dss/data-service-stack.md
alnoki: In the general case I would actually prefer to keep the line included, because theoretically someone could use a DSS container outside the compose environment

However, since per below I've changed to `--force-recreate` I'm fine taking out the forcible container deletion

doc/doc-site/docs/off-chain/dss/data-service-stack.md
alnoki: Added in 94caed2

doc/doc-site/docs/off-chain/dss/data-service-stack.md
alnoki: How about `--force-recreate`, added in 94caed2?

https://docs.docker.com/reference/cli/docker/compose/up

CRBl69: `create` and `build` are two different concepts. Recreate will recreate a new container from the same image, even if there already is a container created from that image and the image hasn't changed. My command will recreate the image to make sure it corresponds to the latest version.

CRBl69 (approved):

PR #732, opened 2024-03-22
[ECO-1451] strip leading zeros in handles

CRBl69:
1. Add a migration to strip leading zeroes from event handles
2. Bump the submodule per https://github.com/econia-labs/aptos-indexer-processors/pull/25


alnoki (changes requested):
- [x] Update submodule

alnoki (approved):

PR #735, opened 2024-03-27
[ECO-1478] update sdk to accept api key

CRBl69:
This PR adds support to add an API key when using the Rust or Python SDK.

- [x] update Rust SDK
- [x] update Python SDK


src/rust/sdk/src/lib.rs
alnoki, line 1: - [ ] Docs on Rust SDK page

alnoki: From eng call:

Add under "structure" section a snippet about adding an auth key, and link to the Aptos developer portal as an example

- [ ] Repeat for Python

alnoki (approved):

PR #736, opened 2024-03-27
[ECO-1276] add suspend/resume runner scripts

CRBl69:
This PR adds two scripts to suspend and resume the runner on GCP to reduce running costs.

### DIFF
--- a/doc/doc-site/docs/off-chain/dss/ci-cd.md
+++ b/doc/doc-site/docs/off-chain/dss/ci-cd.md
@@ -150,11 +150,18 @@ In order to avoid getting charged for a mostly unused service, you can suspend t
 To do so, you can use `scripts/suspend-runner.sh` like so:
 
 ```bash
-source scripts/suspend-runner.sh
+source scripts/suspend-runner.sh GCP_PROJECT_ID
 ```
 
 and `scripts/resume-runner.sh` like so:
 
 ```bash
-source scripts/resume-runner.sh
+source scripts/resume-runner.sh GCP_PROJECT_ID
+```
+
+Note that when resuming, you should wait for the runner to be running before using it.
+You can check its status by running:
+
+```bash
+gcloud compute instances list --filter name=runner
 ```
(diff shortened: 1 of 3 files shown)


alnoki (changes requested):
Docs on terraform docs page?

doc/doc-site/docs/off-chain/dss/ci-cd.md
```diff
@@ -142,3 +142,19 @@ DSS_SOURCE_REV=dss-v1.5.0
...
+To do so, you can use `scripts/suspend-runner.sh` like so:
+
+```bash
+source scripts/suspend-runner.sh
+```
+
+and `scripts/resume-runner.sh` like so:
+
+```bash
+source scripts/resume-runner.sh
```
alnoki, line 159: Now that I'm reading this flow, I'm realizing that a user will get blocked in resuming the runner when the want to re-engage with a project:

Because the `engage-project` script sets the config, which chooses which runner to use, but it also calls the runner and so it'll error out if the runner is paused. I think the `resume-runner` script needs to accept an argument for the project

src/terraform/dss-ci-cd/scripts/resume-runner.sh
```diff
@@ -0,0 +1,2 @@
+gcloud compute instances resume runner
```
alnoki, line 1: How about an optional argument to specify the project for which the runner should be resumed?

Because if you suspend runner A and then `engage-project` with project runner B, you'll need to resume A before you can `engage-project` with it

doc/doc-site/docs/off-chain/dss/ci-cd.md
```diff
@@ -142,3 +142,19 @@ DSS_SOURCE_REV=dss-v1.5.0
...
+```
```
alnoki, line 160: How about a note down here saying that if you resume the runner, give it a minute or two to start up before you re-engage with the project?

alnoki: Oof, it looks like you might then need to make something like `get-runner-startup-logs.sh` but for `get-runner-resumption-logs.sh` which accepts an argument for the project to check in

alnoki (approved):

PR #778, opened 2024-06-17
[ECO-1853] use candlesticks for instant volume

CRBl69:
# Description

This PR fixes the `volume_history` endpoint.

Since the rolling volume pipeline was changed to not create an event when no activity is present on a market, that broke the `volume_history` endpoint.

It was now updated to use candlestick data rather than rolling volume.

# Checklist

- [x] Did you update the changelog?
- [x] Did you check off all checkboxes from the linked Linear task? (Ignore if
  you are not a member of Econia Labs)

### DIFF
--- a/src/rust/dbv2/migrations/2024-06-17-090914_fix_rpc_volume_history/up.sql
+++ b/src/rust/dbv2/migrations/2024-06-17-090914_fix_rpc_volume_history/up.sql
@@ -1,6 +1,24 @@
 -- Your SQL goes here
+
+-- Parameters:
+-- * `market_id`: The market ID that volume is queried for
+-- * `time`: The time up to which volume is measured
+--
+-- Returns:
+-- * `daily`: Market volume in the 24 hours before `time`, measured in indivisible quote subunits
+-- * `total`: All-time market volume before `time`, measured in indivisible quote subunits
 CREATE OR REPLACE FUNCTION api.volume_history (market_id numeric(20,0), "time" timestamptz) RETURNS TABLE(daily NUMERIC(20,0), total NUMERIC(20,0)) AS $$
-    SELECT
-        (SELECT SUM(volume) * (SELECT tick_size FROM market_registration_events WHERE market_id = $1) FROM api.candlesticks WHERE market_id = $1 AND resolution = 60 AND start_time BETWEEN $2 - interval '1 day' AND $2) AS daily,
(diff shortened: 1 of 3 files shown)


src/rust/dbv2/migrations/2024-06-17-090914_fix_rpc_volume_history/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Your SQL goes here
```
alnoki, line 1:
```suggestion
-- Parameters:
-- * `market_id`: The market ID that volume is queried for
-- * `time`: The time up to which volume is measured
--
-- Returns:
-- * `daily`: Market volume in the 24 hours before `time`, measured in indivisible quote subunits
-- * `total`: All-time market volume before `time`, measured in indivisible quote subunits
```

src/rust/dbv2/migrations/2024-06-17-090914_fix_rpc_volume_history/up.sql
```diff
@@ -0,0 +1,6 @@
...
+CREATE OR REPLACE FUNCTION api.volume_history (market_id numeric(20,0), "time" timestamptz) RETURNS TABLE(daily NUMERIC(20,0), total NUMERIC(20,0)) AS $$
+    SELECT
+        (SELECT SUM(volume) * (SELECT tick_size FROM market_registration_events WHERE market_id = $1) FROM api.candlesticks WHERE market_id = $1 AND resolution = 60 AND start_time BETWEEN $2 - interval '1 day' AND $2) AS daily,
```
alnoki, line 4: - [ ] Confirm that `candlesticks` API reports `volume` in ticks
- [ ] Add `COALESCE` to return 0 instead of `null` for strong typing

src/rust/dbv2/migrations/2024-06-17-090914_fix_rpc_volume_history/up.sql
```diff
@@ -0,0 +1,6 @@
...
+        (SELECT SUM(volume) * (SELECT tick_size FROM market_registration_events WHERE market_id = $1) FROM api.candlesticks WHERE market_id = $1 AND resolution = 60 AND start_time < $2) AS total;
```
alnoki, line 5: - [ ] Add `COALESCE` to return 0 instead of `null` for strong typing

alnoki (changes requested):
- [x] Add changelog entry

alnoki (approved):

PR #792, opened 2024-07-24
[ECO-1997] Add transaction version in MQTT price levels topic payload

CRBl69:
# Description

This PR adds the transaction version at which the price levels data was
calculated in the MQTT `levels` topic payload.

This will help clients generate an accurate representation of data using the
REST API, then stay up to date using an MQTT subscription, without counting any
events twice, or missing any events.

# Testing

Spin up a DSS.

Subscribe to `levels/#`.

# Checklist

- [x] Did you check off all checkboxes from the linked Linear task? (Ignore if
  you are not a member of Econia Labs)


src/rust/mqtt-publisher/src/main.rs
```diff
@@ -136,16 +137,26 @@ async fn eventpoll_loop(mut eventloop: EventLoop) -> Result<()> {
...
+    txn_version: u128,
 }
 
 async fn price_level_loop(db_url: &str, mqtt_client: Arc<RwLock<AsyncClient>>) -> Result<()> {
     let pool = PgPool::connect(&db_url).await?;
 
     loop {
         tokio::time::sleep(Duration::from_millis(50)).await;
+        let mut tx = pool.begin().await?;
+        tx.execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;").await?;
```
alnoki, line 149: Do we already have a helper for repeatable read?

If so I suggest using that so that we are using consistent abstraction

src/rust/mqtt-publisher/sqlx_queries/get_txn_version.sql
alnoki, line 1: How about renaming this to `get_user_history_last_indexed_txn`?

Since there is a different txn version for the latest processor insert, for example

src/rust/mqtt-publisher/src/main.rs
```diff
@@ -136,16 +137,26 @@ async fn eventpoll_loop(mut eventloop: EventLoop) -> Result<()> {
 struct PriceLevel {
     price: u128,
     size: u128,
+    txn_version: u128,
```
alnoki, line 140: - [ ] Update `mqtt` docs site page to reflect the new payload
- [ ] Also mention under the price levels entry on the docs site that you'll need to sett the `MQTT_PRICE_LEVELS` environment variable (it is already in the `example.env` file, of course, but who have set up deployments and then later wanted price levels didn't think to check): here you can link back to the DSS landing page and talk about when you copy the `exmaple.env` file for example

alnoki (changes requested):

src/rust/mqtt-publisher/src/main.rs
CRBl69: We have a helper function in the aggregator, so I can't use it here. One way to do this would be to create a new crate with just this function in it. I think this would just make everything more complex IMO.

src/rust/mqtt-publisher/sqlx_queries/get_txn_version.sql
CRBl69: Done

alnoki (approved):

PR #808, opened 2025-04-19
Migrate processor to indexer SDK

CRBl69:
# Description

This PR updates the DSS processor to use the newly released Aptos indexer SDK.

### DIFF
--- a/src/docker/processor/start.sh
+++ b/src/docker/processor/start.sh
@@ -10,15 +10,12 @@ rm /app/out.csv
 
 echo "health_check_port: 8085
 server_config:
-  processor_config:
-    type: econia_transaction_processor
-    econia_address: $ECONIA_ADDRESS
-  postgres_connection_string: $DATABASE_URL
-  indexer_grpc_data_service_address: $GRPC_DATA_SERVICE_URL
-  indexer_grpc_http2_ping_interval_in_secs: 60
-  indexer_grpc_http2_ping_timeout_in_secs: 10
-  auth_token: $GRPC_AUTH_TOKEN
-  number_concurrent_processing_tasks: 1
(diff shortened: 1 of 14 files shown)


src/docker/processor/start.sh
```diff
@@ -10,15 +10,12 @@ rm /app/out.csv
...
 server_config:
-  processor_config:
-    type: econia_transaction_processor
-    econia_address: $ECONIA_ADDRESS
-  postgres_connection_string: $DATABASE_URL
-  indexer_grpc_data_service_address: $GRPC_DATA_SERVICE_URL
-  indexer_grpc_http2_ping_interval_in_secs: 60
-  indexer_grpc_http2_ping_timeout_in_secs: 10
-  auth_token: $GRPC_AUTH_TOKEN
-  number_concurrent_processing_tasks: 1
```
alnoki, line 21: I think this line is still needed to prevent out of order behavior

CRBl69: Actually, I think transactions are processed in order, but one batch is not awaited to be processed before the next one starts to process.

So there might be one case where this is a problem:

Transaction batch A comes in with 1000 transactions and 10000 econia events, takes a while to insert.

Transaction batch B comes in with 1000 transactions but only 1 econia event. Inserts very fast.

If the processor does not wait for the first batch processing to be finished, and the second starts and finishes before the first, this could cause an issue.

Noting that this is not a problem in the emojicoin.fun processor, as that one uses a rwlock that forbids this from happening (for unrelated reasons). If Aptos Labs confirms that processing of batch B can start before processing of batch A has ended, I will update this to include a rwlock as well.

CRBl69: I will actually include a rwlock in order to not have to await Aptos Labs response, since there are no real dowsides to it

